PR #1, opened 2015-09-30
GeoJSON support with default CRS only

AkeluX:
Add a new format support, GeoJSON. CRS is currently limited to urn:ogc:def:crs:OGC:1.3:CRS84. This will be soon improved using Proj4js.


tag3:
Would you please remove .idea/workspace.xml from the request. That is a per-person file.

pdavidc:
Servet, Would you please change the name of the top level class from GeoJSON to GeoJSONParser in order to make it more descriptive?

servetg:
GeoJSON renamed GeoJSONParser, .idea/workspace.xml removed, named and linked CRS support added.

PR #3, opened 2015-10-05
Feature/kml support

balhar-jakub:
Support for KML. 
At currrent state it should be able to parse the document fully. It should also be able to render most of the geometries and placemarks. And apply at least to some extent styling.


tag3:
I tried running examples/KmlSupport but got an error. It looks like the KML file constructor is expecting an object but is being passed a text string (the KML).

balhar-jakub:
You are right. I updated the kmlfile to support remote files as well as
passed strings but forgot the example.

Thank you for reminder.
Dne 5. 10. 2015 19:47 napsal uživatel "Tom Gaskins" <
notifications@github.com>:

PR #5, opened 2015-10-08
Demo/analytical surface

servetg:
Added AnalyticalSurface demo and dataset


tag3:
The rule limiting use of external libraries applies only to the core code (that in src). Examples are free to use external libraries. In the case of this pull request, d3.js is included in a dependencies directory. If possible, it would be better for the example to refer to it at a CDN, perhaps https://cdnjs.cloudflare.com/ajax/libs/d3/3.5.6/d3.min.js.

servetg:
I've removed the dependencies directory and changed to CDN reference.

PR #7, opened 2015-10-16
Feature/annotations

revo09:


tag3:
The implementation looks great! Thanks! The one change we'd like to request is to rename the example  program Annotations.html rather than AnnotationsExample.html. Nice work!

revo09:
The final fixes have been made so I believe it's ok for a merge.

PR #9, opened 2015-11-30
Feature/kml support review

balhar-jakub:
Initial support for KML file format. 
It brings support for parsing all elements of KML document and rendering following elements:
- Placemark
- Polygon
- Point
- LineString
- LinearRing
- PolyStyle
- LineStyle
- MultiGeometry
- Style
- Document - Very crude support so far
- Folder - Very crude support so far


tag3:
jakub, Thanks for the pull request. However, I notice that some out-of-date core files are included. An example is ElevationModel.js. The pull request appears to be pushing a change that eliminates a previous change. There are several such files. Is it possible that your source wasn't updated before the pull request?

 I also notice the use of extend rather than our standard pattern of:
<constructor>.prototype = Object.create(<superclass>.prototype);
I think that's what extend is trying to do.

balhar-jakub:
The reason for using extend is for the object like Placemarks and Geometry objects, which have internal representation in the framework KmlPlacemark should be descendant of Placemark, while at the same time they should use the methods from the KmlObject. 
extend method is used to mix in the methods from the KmlObject. Probably it should be names mixin instead. 

With the source, I was merging it with our default branch after the pull request from your master branch. I am not really sure how this happened.

tag3:
Jakub,

Thanks for the explanation of extend.

As to the old included source, I don’t know how that happened or could have happened. But for instance the examples/Shapefiles.js file I have most recently checked out has a @version string with a date of 2015-07-31, whereas the one the pull request seeks to check in has a @version string date of 2015-07-15, and the diff at Github shows that it’s replacing that more recent version string. This will need to be figured out and rectified before the pull request can be integrated.

Thanks
Tom

balhar-jakub:
I found the cause. It was in the initial commit of the KML Support, which is why it took me so much time to find it. I incorrectly merged following files, commit I am reffering to is 667fe1dc5442eef5e9eade79ed9b6648a546c8bd:

ArgumentError.js
BingWMSLayer.js
BMNGLandsatLayer.js
BMNGLayer.js
EarthRestElevationModel.js
ElevationModel.js
Layer.js
Logger.js
Shapefile.js
Shapefiles.js
Texture.js
TiledImageLayer.js
WmsLayer.js
WmsUrlBuilder.js
WWUtil.js

To fix this I created patches, which should be valid against current state. The most complex merges were done for Layer.js and TiledImageLayer.js
Please take a look whether I was right.

PR #12, opened 2016-01-20
Fix/surface shapes color

strikerM:


pdavidc:
Florin, Thanks for investigating and sending this pull request. We've merged your changes in to the main branch.

PR #25, opened 2016-05-18
Callback GeoJSON

GabrielePrestifilippo:
Inserted the callback in the GeoJSON load


pdavidc:
Gabriele, The proposed GeoJSON parser completion callback looks sound, but it needs to be modified to match the pattern found in Shapefile. Please review Shapefile's parserCompletionCallback and match that pattern here in GeoJSON. In particular, we're interested in achieving consistent naming, argument order, object properties, and documentation.

AkeluX:
Hello Gabriele, the documentation/comments say "GeoJSON shapefiles" now. I guess this was copy/pasted from the shapefile implementation. I would write "GeoJSON" only.

GabrielePrestifilippo:
Thank you @AkeluX, I corrected the syntax. It is indeed a shape, but maybe writing shapefile could have been misleading.

pdavidc:
Gabriele, Thank you for making these changes. This is closer to the Shapefile interface, but not quite what we're looking for. In particular the argument order and property order for the callbacks are different. Our goal is to establish very consistent patterns that enable application developers to learn a pattern one place and use it elsewhere. Changing argument order forces one to re-learn a similar pattern each place.

pdavidc:
Gabriele, The latest changes to this pull request look perfect. Thank you.

PR #27, opened 2016-06-06
Support for Region, ScreenOverlay, GroundOverlay

balhar-jakub:
The KmlFile is now Renderable itself and on top of that it contains basic support for Region, ScreenOverlay and GroundOverlay elements.


pdavidc:
This pull request looks great. Thanks Jakub.

PR #28, opened 2016-06-07
Atmosphere and day/night effect

strikerM:


pdavidc:
Florin, We've reviewed this pull request and have identified the changes needed before this feature may be merged. The changes are indicated by inline comments in the source.

src/layer/AtmosphereLayer.js
```diff
@@ -0,0 +1,414 @@
...
+         * @classdesc Provides a layer showing the Earth's atmosphere.
+         * @param {URL} nightImageSource optional url for the night texture.
+         * @augments Layer
+         */
+        var AtmosphereLayer = function (nightImageSource) {
+
+            Layer.call(this, "Atmosphere");
+
+            //image for the night texture
+            this._nightImageSource = nightImageSource || '../images/dnb_land_ocean_ice_2012.png';
```
pdavidc: The base path for the default image must be WorldWind.configuration.baseUrl + "images/dnb_land_ocean_ice_2012.png"

See BMNGOneImageLayer for an example.

src/layer/AtmosphereLayer.js
```diff
@@ -0,0 +1,414 @@
...
+             */
+            nightImageSource: {
+                get: function () {
+                    return this._nightImageSource;
+                },
+                set: function (value) {
+                    this._nightImageSource = value;
+                }
+            },
+
```
pdavidc: The properties skyWidth, skyHeight, skyPoints and skyTriStrip are considered internal implementation details. Please mark them as internal in the property declarations above, and remove their associated property accessors.

examples/Atmosphere.js
```diff
@@ -0,0 +1,61 @@
...
+            {layer: new WorldWind.CompassLayer(), enabled: false},
+            {layer: new WorldWind.CoordinatesDisplayLayer(wwd), enabled: false},
+            {layer: new WorldWind.ViewControlsLayer(wwd), enabled: false}
+        ];
+
+        for (var l = 0; l < layers.length; l++) {
+            layers[l].layer.enabled = layers[l].enabled;
+            wwd.addLayer(layers[l].layer);
+        }
+
```
pdavidc: It would be better if the example demonstrated usage of the default image. Please remove specification of the night image here.

src/layer/AtmosphereLayer.js
```diff
@@ -0,0 +1,414 @@
...
+            var gl = dc.currentGlContext,
+                program = dc.currentProgram,
+                skyData = this.skyData,
+                vboId;
+
+            if (!skyData.verticesVboCacheKey) {
+                skyData.verticesVboCacheKey = dc.gpuResourceCache.generateCacheKey();
+            }
+
```
pdavidc: The adaptation of sky vertex buffer objects from WWAndroid to WebWW is generally quite good. However, I don't see a need for the refreshVertexBuffer and refreshIndices buffer properties. It would be better if the buffer were created and its data specified in one step.

src/layer/AtmosphereLayer.js
```diff
@@ -0,0 +1,414 @@
...
+                skyData.refreshVertexBuffer = true;
+            }
+
+            gl.bindBuffer(gl.ARRAY_BUFFER, vboId);
+            if (skyData.refreshVertexBuffer) {
+                gl.bufferData(gl.ARRAY_BUFFER, new Float32Array(this.skyPoints), gl.STATIC_DRAW);
+                dc.frameStatistics.incrementVboLoadCount(1);
+                skyData.refreshVertexBuffer = false;
+            }
+
```
pdavidc: I don't see where SkyProgram specifies the vertexPointLocation attribute. Additionally, the emerging pattern in WebWW is to specify vertex attribute locations explicitly at GLSL compilation time, then hard code the specified locations in our GL calls. In this case, program.vertexPointLocation can be replaced with 0.

src/layer/AtmosphereLayer.js
```diff
@@ -0,0 +1,414 @@
...
+        AtmosphereLayer.prototype.applySkyIndices = function (dc) {
+
+            var gl = dc.currentGlContext,
+                skyData = this.skyData,
+                vboId;
+
+            if (!skyData.indicesVboCacheKey) {
+                skyData.indicesVboCacheKey = dc.gpuResourceCache.generateCacheKey();
+            }
+
```
pdavidc: See the above comment regarding VBO construction and data specification.

src/layer/AtmosphereLayer.js
```diff
@@ -0,0 +1,414 @@
...
+            gl.depthMask(false);
+            gl.frontFace(gl.CW);
+
+            this.setSkyPoints(dc, program.getAltitude());
+            this.setSkyTrianglesIndices();
+
+            this.applySkyVertices(dc);
+            this.applySkyIndices(dc);
+
```
pdavidc: This is the place to call enableVertexAttribArray and disableVertexAttribArray, grouped with the other direct GL state specification (depthMask and frontFace).

src/layer/AtmosphereLayer.js
```diff
@@ -0,0 +1,414 @@
...
+            // Use this layer's night image when the light location is different than the eye location.
+            if (this.nightImageSource && this.lightLocation) {
+                this._activeTexture = dc.gpuResourceCache.resourceForKey(this.nightImageSource);
+                if (!this._activeTexture) {
+                    this._activeTexture = dc.gpuResourceCache.retrieveTexture(gl, this.nightImageSource);
+                }
+                textureBound = this._activeTexture && this._activeTexture.bind(dc);
+            }
+
```
pdavidc: It's not obvious, but this code accesses an internal property of the Tessellator, currentTiles, which could be empty if the tessellator were modified to avoid keeping tile references post-tessellation. Instead of terrain.tessellator.currentTiles, use terrain.surfaceGeometry, which is also a list of TerrainTile instances.

src/layer/AtmosphereLayer.js
```diff
@@ -0,0 +1,414 @@
...
+                // Use a tex coord matrix that registers the night texture correctly on each terrain.
+                if (textureBound) {
+                    this._texMatrix.setToUnitYFlip();
+                    this._texMatrix.multiplyByTileTransform(currentTile.sector, this._fullSphereSector);
+                    program.loadTexMatrix(gl, this._texMatrix);
+                }
+
+                // Draw the tile, multiplying the current fragment color by the program's secondary color.
+                program.loadFragMode(gl, program.FRAGMODE_GROUND_SECONDARY);
+                gl.blendFunc(gl.DST_COLOR, gl.ZERO);
```
pdavidc: Use terrain.beginRendering instead of terrain.tessellator.beginRendering. Apply that pattern to endRendering as well.

src/layer/AtmosphereLayer.js
```diff
@@ -0,0 +1,414 @@
...
+                textureBound;
+
+            program.loadGlobeRadius(gl, dc.globe.equatorialRadius);
+
+            program.loadEyePoint(gl, dc.navigatorState.eyePoint);
+
+            program.loadLightDirection(gl, this.activeLightDirection);
+
+            program.setScale(gl);
+
```
pdavidc: It appears that calls to enableVertexAttribArray for the vertexPoint and vertexTexCoord attributes are missing here. As mentioned above, specify these vertex attribute locations explicitly at GLSL compilation time, then hard code the specified locations in our GL calls. In this case, program.vertexPointLocation can be replaced with 0 and program.vertexTexCoordLocation can be replaced with 1.

src/layer/AtmosphereLayer.js
```diff
@@ -0,0 +1,414 @@
...
+                gl.blendFunc(gl.ONE, gl.ONE);
+                terrain.tessellator.beginRendering(dc);
+                terrain.beginRenderingTile(dc, currentTile);
+                terrain.renderTile(dc, currentTile);
+                terrain.endRenderingTile(dc, currentTile);
+                terrain.tessellator.endRendering(dc);
+            }
+
+            // Restore the default World Wind OpenGL state.
+            gl.blendFunc(gl.ONE, gl.ONE_MINUS_SRC_ALPHA);
```
pdavidc: See the comment above regarding program.vertexPointLocation and program.vertexTexCoordLocation.

src/layer/AtmosphereLayer.js
```diff
@@ -0,0 +1,414 @@
...
+            var result = [];
+            var index = [];
+            var vertex = 0;
+
+            for (var latIndex = 0; latIndex < numLat - 1; latIndex++) {
+                // Create a triangle strip joining each adjacent column of vertices, starting in the bottom left corner and
+                // proceeding to the right. The first vertex starts with the left row of vertices and moves right to create
+                // a counterclockwise winding order.
+                for (var lonIndex = 0; lonIndex < numLon; lonIndex++) {
+                    vertex = lonIndex + latIndex * numLon;
```
pdavidc: The usage of a separate index array here is a pattern that was useful in WWAndroid, but is not necessary in WebWW. The 'index' array usage can be removed, and values pushed directly in the 'result' array.

src/layer/AtmosphereLayer.js
```diff
@@ -0,0 +1,414 @@
...
+
+            program.loadModelviewProjection(gl, dc.navigatorState.modelviewProjection);
+
+            program.loadLightDirection(gl, this.activeLightDirection);
+
+            program.setScale(gl);
+
+            gl.depthMask(false);
+            gl.frontFace(gl.CW);
+
```
pdavidc: The sky buffer objects for vertices and elements are only used for rendering. Given that, it's unnecessary to keep them around here in client memory. Please remove the class properties for these buffer objects, and simply compute and store them as needed directly into the vertex buffer objects. The result should be applySkyVertices and applySkyIndices.

src/shaders/AtmosphereProgram.js
```diff
@@ -0,0 +1,322 @@
...
+         * @param {WebGLRenderingContext} gl The current WebGL context.
+         * @param {Vec3} vector The vector to load.
+         * @throws {ArgumentError} If the specified vector is null or undefined.
+         */
+        AtmosphereProgram.prototype.loadVertexOrigin = function (gl, vector) {
+            if (!vector) {
+                throw new ArgumentError(
+                    Logger.logMessage(Logger.LEVEL_SEVERE, "AtmosphereProgram", "loadVertexOrigin", "missingVector"));
+            }
+
```
pdavidc: It would be better to use uniform3f here instead of uniform3fv, thereby removing an unnecessary array allocation. Please apply that pattern to the remaining uniform code in AtmosphereProgram.

src/shaders/AtmosphereProgram.js
```diff
@@ -0,0 +1,322 @@
...
+        AtmosphereProgram.prototype = Object.create(GpuProgram.prototype);
+
+        /**
+         * Returns the atmosphere's altitude.
+         * @returns {Number} The atmosphere's altitude in meters.
+         */
+        AtmosphereProgram.prototype.getAltitude = function () {
+            return this.altitude;
+        };
+
```
pdavidc: I don't see a reason to have rayleighScaleDepth exposes as a public property. Assuming there isn't one, please remove this method and replace its usage with direct access to rayleighScaleDepth.

src/navigate/Navigator.js
```diff
@@ -112,6 +112,12 @@ define([
             // modelview matrix and before computing the near clip distance. The far clip distance depends on the
             // modelview matrix, and the near clip distance depends on the far clip distance.
             this.farDistance = WWMath.horizonDistanceForGlobeRadius(globeRadius, eyePos.altitude);
+
```
pdavidc: We discussed the NASA team handling the changes necessary to support a far clip distance that doesn't obscure the horizon. Adding a large constant here will cause more shapes and tiles to be considered each frame, especially when the viewer is close to the ground. We'll need to find another way to handle this change.

One possibility is for the ESA team to implement the far clip convention found in WWAndroid's source in WorldWindow.computeViewingTransform, and test the resultant performance difference. In particular, we need to quantify the cost of increasing the far clip distance on terrain tile count and image tile count.

PR #30, opened 2016-06-29
Fix blue marble layer

balhar-jakub:
Fix Blue marble layer example to work in local environment.


pdavidc:
I think there's still some confusion about this layer's purpose. It's designed to read data from the local disk. While that doesn't work until one establishes a local copy of the Blue Marble data, we also don't want to point at data stored at worldwindserver.net. We either need to leave this alone or provide documentation clarifying the intended operation.

balhar-jakub:
Hi,

Ok. Let's close this request and I will simply update the documentation and
then create new one.

Best regards,

Jakub

PR #33, opened 2016-07-05
OGC TMS implementation

thales-geo:
Child of #528 

This is the feature branch for the TMS protocol implementation. It includes both code and examples. Demos are available here: https://thales-geo.github.io/webworldwind-demos/


pdavidc:
We've recently completed an architectural review of WebWorldWind's image layers. We found that the current pattern of defining a new layer for each service type (WMS, WMTS, TMS, Bing, etc.) needs to change. We're closing this pull request in anticipation of a implementing TMS with a new approach.

The current approach used in the layers for WMS, WMTS, Bing, and the proposed TMS in this pull request rely on either inheritance of TiledImageLayer, or rely on what amounts to a copy and paste of TiledImageLayer, with the necessary adjustments for that service. We've found this approach to be difficult to maintain, and difficult to understand. Enhancements and bug fixes that apply to all image layers are much more difficult to implement, and the extra work required to understand the implications of those changes across similar but different implementations is burdensome. 

Furthermore, the relationship of WorldWind's internal image tiling structure to performance is too important to leave the tiling scheme up to the image provider. The layer must define the tiling scheme, and the imagery must adapt to that scheme. 

Given these discoveries, it is our intention to build a common image layer that can be _composed_ with an image provider/retriever to supply tiles to the WebWorldWind client. The layer's management of image tiles and WebGL resources will remain consistent across all image service types. Each image service will supply the glue that enables services like WMS, WMTS, Bing, TMS, etc. to create image tiles appropriate for use in WebWorldWind.

See epic #528 for future TMS work
See epic #568 for the architectural changes needed to build a common image layer

PR #41, opened 2016-07-27
Karma Tests

AkeluX:
There are still 7 KML tests to be ported to Jasmine. They are currently commented out and marked with a TODO. Nonetheless, this should get us started with the new test framework.


test/formats/kml/util/ImagePyramid.test.js
```diff
@@ -2,20 +2,32 @@
...
+            if (this.index > 0)
+            {   var failed = jsApiReporter.specResults(this.index -1, 1)[0].failedExpectations;
+                console.log('failed: ', failed);
+                if (failed.length > 0)
+                {
+                    console.log('After: ', this, failed[0].message);
+                    alert('ha');
+                }
+            }
+        });
```
balhar-jakub: This part works only with one test. When there are multiple tests it fails with exception for undefined. So if we need this mechanism, we need to update it somehow.

balhar-jakub:
Remaining tests were ported.

Issue #51, opened 2016-09-18
mouseover on collada objects


strangerintheq:
sometimes mouseover on one collada object is recognized as mouseover over other collada object

strikerM:
The issues should be fixed now. See pull request #53 

@strangerintheq when you get the chance, would you please verify that this issue is resolved?

strangerintheq:
perfect

PR #53, opened 2016-09-21
bugfix collada model picking

strikerM:
In response to #51


src/formats/collada/ColladaScene.js
```diff
@@ -809,7 +809,7 @@ define([
             var gl = dc.currentGlContext,
                 program = dc.currentProgram;
 
-            if (hasLighting) {
+            if (hasLighting || dc.pickingMode) {
```
strikerM: Yes, you are right. It should be (hasLighting && !dc.pickingMode)

PR #60, opened 2016-10-04
Skybox imlementation

strikerM:


AkeluX:
This will ultimately be replaced by the new star field layer when it's ready. I suggest to reject and close. Any objections?

AkeluX:
Will be replaced by the new star field

PR #73, opened 2017-01-10
Update BingImageryUrlBuilder.js

YvesE:
Correction of the "Bing maps key not send in the URL" issue.


AkeluX:
@YvesE, thank you for contributing a solution to this bug. We would like to merge it, but believe that removing the nesting of the conditionals would make it easier to comprehend. Would you be willing to change it for something along the line of this pseudo code?

```javascript
// Use key specified for this layer
this.bingMapsKey = bingMapsKey;

// If none, fallback to key specified globally
if (!this.bingMapsKey) {
    this.bingMapsKey = WorldWind.BingMapsKey;
}

// If none, fallback to default demo key
if (!this.bingMapsKey) {
    this.bingMapsKey = wwBingMapsKey;
    BingImageryUrlBuilder.showBingMapsKeyWarning();
}
```

YvesE:
Removed the nesting of the conditionals as suggested.

pdavidc:
@Beak-man It appears to me that @YvesE has made the changes we requested. Would you please verify whether or not that's the case? If the changes look good, please add your review.

Issue #77, opened 2017-02-09
KML`s tag Pair   ,getStyle is wrong


QingMings:
[Pair](https://github.com/NASAWorldWind/WebWorldWind/blob/master/src/formats/kml/util/Pair.js)
the line 94 
    `   StyleResolver `  Should not be used directly，
Here is KML for testing
 [huoche.kml](https://github.com/QingMings/kmlupload/blob/master/huoche.kml)

balhar-jakub:
If I understand it correctly, the highlight style isn't applied correctly? Is there anything else?
I will take a look at possible fix.

QingMings:
yes , the Pair tag  not correct analysis, hope  you fix it ,thanks :)

QingMings:
@balhar-jakub  
How to know a KML file rendering complete,
in my code : 
`kmls.getKmlData().then(function(kmlFile){`    
                 ```          var renderableLayer = new WorldWind.RenderableLayer('kmlLayer')```
                  `         renderableLayer.addRenderable(kmlFile);`
                  `          kmls.wwd.addLayer(renderableLayer);`
                 `          kmls.wwd.redraw();`
`})`

when `wwd.redraw()`,  `kmlFile` is  not rendering complete,so the screen is blank ,Until I drag the mouse, trigger `redraw`

balhar-jakub:
It seems to me that what you want to get is the moment when you can call redraw on the globe so that the KML is displayed. 

The KmlFile constructor returns Promise. When this promise is resolved, the KmlFile is ready for the rendering. If at this moment you call redraw on the WorldWindow it should correctly display the KML file. There is an example in the KmlExample.js in the examples section. This is a simplified version which simply adds the KML File. 

    var kmlFilePromise = new KmlFile('data/KML_Samples.kml', []);
    kmlFilePromise.then(function (kmlFile) {
            var renderableLayer = new WorldWind.RenderableLayer("KML file shapes");
            renderableLayer.addRenderable(kmlFile);

            wwd.addLayer(renderableLayer);
            wwd.redraw();
        });

I am not sure what getKmlData() call does. Could you please provide this piece of the code? Maybe there is something going on in there, that I am missing.

QingMings:
![image](https://cloud.githubusercontent.com/assets/22107668/24100250/5c229736-0dae-11e7-9041-1dbdcc88c576.png)


just  pass parameter  ,return `new KmlFile`。
thanks .

[...]

PR #80, opened 2017-02-16
fix  Pair tag   getStyle()   execution failure 

QingMings:
fix  Pair tag   getStyle()   execution failure  ，see issus #77

### DIFF
--- a/src/formats/kml/KmlFile.js
+++ b/src/formats/kml/KmlFile.js
@@ -50,6 +50,7 @@ define([
      * @alias KmlFile
      * @classdesc Support for Kml File parsing and display.
      * @augments KmlObject
+     * 2017-02-16   fix  Kmz internal KML suffix get  failed
      */
     var KmlFile = function (url, controls) {
         var self = this;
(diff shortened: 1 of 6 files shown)


balhar-jakub:
The changes to the Style resolver seems fine to me and all changes associated to this: KmlStyleMap, Pair, StyleResolver. 
What I don't understand is why the change in the jszip.js What is the reason behind this change. The rest seems ok for the merge, but I need to understand this one. Could you provide some example of file, which works incorrectly without this change?

QingMings:
In   `KmlFile.js`  ，
The 83 line.

83  `         kmzFile.files.forEach(function (file) {   `
 84`                          if (file.endsWith(".kml") && rootDocument == null) {`
  85`                             if (file.name.endsWith(".kml") && rootDocument == null) {`
 86   `                              rootDocument = file.asText();`
   87`                          }`
    88 `                   });`

The  `kmlFile.files`  is  a object  ,not a  array , can‘t use   `forEach` method .
 So i change the `jszip.js`   ，  let it return a Array .
Maybe I'm not doing it right. Just to solve the problem, without considering the rationality.
Sorry for this .

balhar-jakub:
I missed this one. Don't change the jszip. Instead this code should be changed. It should be working if it is changed to:

    Object.keys(kmzFile.files).forEach(function(key){
        var file = kmzFile.files[key];
    });

Retrieve all the relevant files and then manage the previous function on each of them. Is it obvious what I am trying to explain?

QingMings:
Maybe you can help me with this  #84 ，

QingMings:
yes, I understand，I will correct it。:)

PR #86, opened 2017-02-22
Add starfield layer

strikerM:


AkeluX:
This isn't directly for the starfield, but now that we include an util for computing the position of the sun, we could let the atmosphere layer react to its time property as well. What do you think?

pdavidc:
Yann, I think that's a good idea, and the CelestialProjection class is a good start. I'm unsure of this class' name and place within the SDK. By name and some methods CelestialProjection seems related to GeographicProjection, but CelestialProjection is not a map projection and includes some utilities that appear more general. Would you be willing to review this class on the basis of its naming and its cohesion?

AkeluX:
Absolutely. I'll create an issue to track this idea of changing the Atmosphere layer a bit. We can first finalise the Starfield and this new util. Some reviews from my side are also pending here by the way.

AkeluX:
The goal of CelestialProjection currently boils down to computing the sun's position in geographic and celestial coordinates. celestialToGeographic, computeJulianDate and normalizeAngle are actually helpers to reach this purpose. I would thus remove them from the public interface and have computeSunCelestialLocation take a regular date as argument. This should help with the cohesion from the perspective of the public interface.

I haven’t found such normalization of angles and conversions (celestialToGeographic and Julian date) elsewhere in the SDK so they could stay in this class in my opinion.

Regarding the class name, I would have it relate to its core purpose. Maybe something like SunPosition with methods like getAsGeographicLocation and getAsCelestialLocation.

Regarding the location of this class in the SDK, I don’t see a more appropriate place than the very generic util, at the moment.

@strikerM @pdavidc what do you think?

strikerM:
CelestialProjection was intended to be used for any celestial body.

At the moment CelestialProjection is only useful for the Sun position, but it was originally implemented with the Moon as well and had the methods computeMoonCelestialLocation and computeMoonGeographicLocation.

The methods celestialToGeographic, computeJulianDate and normalizeAngle are helpers that can be used for any celestial body.

My reasoning was that if we want to add a new celestial body, it would only require a computeXCelestialLocation method for that particular object.

All celestial bodies (stars, Sun, Moon) require converting to julianDate.
The reason computeSunCelestialLocation takes a julianDate instead of a javascript Date was so that this conversion is done only once and then reused.

There is a computeSunGeographicLocation method that takes a javascript Date are returns LatLon.

Let me know if you want me to change anything.

strikerM:
I refactored CelestialProjection, let me know what you think.
@AkeluX @pdavidc

AkeluX:
For me, it's fine.

src/shaders/StarFieldProgram.js
```diff
@@ -0,0 +1,232 @@
...
+                    'uniform vec2 magnitudeRange;\n' +
+
+                    'varying float magnitudeWeight;\n' +
+
+                    //normalizes an angle between 0.0 and 359.0
+                    'float normalizeAngle(float angle) {\n' +
+                    '   float angleDivisions = angle / 360.0;\n' +
+                    '   return 360.0 * (angleDivisions - floor(angleDivisions));\n' +
+                    '}\n' +
+
```
pdavidc: @strikerM, could this be replaced by the built-in GLSL step function?
http://docs.gl/el3/step

src/util/SunPosition.js
```diff
@@ -0,0 +1,173 @@
...
+                    month += 12;
+                }
+
+                var A = Math.floor(year / 100);
+                var B = 2 - A + Math.floor(A / 4);
+                var JD0h = Math.floor(365.25 * (year + 4716)) + Math.floor(30.6001 * (month + 1)) + day + B - 1524.5;
+
+                return JD0h + dayFraction;
+            },
+
```
pdavidc: @strikerM, This function looks appropriate for WWMath. Would you please move it there? Also, if you'd compare it to Android's WWMath.normalizeAngle we need to ensure that the two are equivalent.

strikerM:
@pdavidc 
I've added normalizeAngle360, from WWAndroid, to WWMath and replaced the if_gt function with the built in step function.

AkeluX:
@strikerM with your latest commit, it seems you've removed the test associated with the normalizeAngle360 logic. Could you create a WWMath.test.js and add your test there instead?

strikerM:
I've added the test.

PR #93, opened 2017-03-07
Enable configurable limits to navigation

PHPaul-ES:
Child of #508 

Setting some fixed values as variables, to enable a bit of extra customization in the navigation
Able to limit navigation to a sector, zoom and tilt limits.
Should navigate as usual, unless the new attributes are changed


pdavidc:
@PHPaul-ES Thank you for your contribution. We're overhauling WorldWind's navigation system, and will take the underlying use case into consideration in our ongoing work on navigation.

pdavidc:
The navigation architecture is changing via #279, rendering this change obsolete.

PR #101, opened 2017-03-31
Improve the navigator architecture (camera and custom controllers)

AkeluX:
This pull request was created so we can start finalising this. It's a lot of changes and I'm not familiar with the corresponding code in the Android version.


AkeluX:
@balhar-jakub couple of questions to start with

Are the changes in .idea required? Do they bring anything?

Is this change 100% backward compatible?

balhar-jakub:
They weren't required and as such I already removed these changes.

AkeluX:
Isn't this still missing the conversion methods for the 2D projections? You wanted to look into this @pdavidc, didn't you?

AkeluX:
@balhar-jakub I just realised that this pull request was for a merge into master. I've changed it to develop. As you can see, there are some conflicts. They are easy enough to solve, but then there are other tests failing for some reason. Could you update 'feature/camera' with the latest state of develop, solve these conflicts and verify that everything still runs fine? This will make @pdavidc's review much easier.

pdavidc:
@AkeluX Thanks for catching this. Resolving these conflicts is indeed a necessary step in order to finish the camera work @balhar-jakub started.

@balhar-jakub Let me know whether or not you're able to resolve of these conflicts. I can certainly handle them if you're too busy.

balhar-jakub:
I went through the code and merged the develop branch into the camera branch and fixed the tests.

pdavidc:
Thank you @balhar-jakub. Starting today, I'll be reviewing this code and completing the (intentionally) left out portions dealing with lookAt/camera and 2D projections. My apologies for the long delay in getting to this.

I'll also take the opportunity to bringing this up to date with the current Android camera architecture, which we adjusted based on your feedback. In particular, the lookAt/camera conversions are no longer on Globe.

Also, I'll likely break this up into a series of smaller compartmentalized changes and pull requests in order to make it easier for our teams to review and incorporate.

AkeluX:
@pdavidc, we've discovered that there is an issue in this branch with the GoToAnimator when the globe is tilted. @balhar-jakub is taking a look at this and will report back here. This might be good inputs for your own changes.

pdavidc:
Thanks for letting me know.

balhar-jakub:
I went through the code and first to reproduce the bug, it is enough to set heading and tilt to 15 degrees, but it won't be the only problematic use case. To see what it actually does, it is enough to put this code at the end of the GoToLocation.js example

        var lookAt = new WorldWind.LookAt();
        lookAt = wwd.navigator.getAsLookAt(wwd.globe, lookAt);
        lookAt.heading = 15;
        lookAt.tilt = 15;
        wwd.navigator.setAsLookAt(wwd.globe, lookAt);
        wwd.redraw();

        setTimeout(function(){
            wwd.goTo(new WorldWind.Location(50, 14));
        }, 5000);

There are two cases. First is the one that works exactly as it works now. The GoToAnimator calculates distance per tick and every tick moves the globe in given distance. In the last tick, it gets to a point where the difference in meters is less than one. 

The second case differs in the last tick. It updates the latitude and longitude but due to issues around rounding the distance between target point the calculated distance is more than one meter. This happens partially because when setting the state of the navigator as LookAt it is actually internally transformed to Camera, which means that there is some math performed and in certain cases, the rounding errors add up enough to actually differentiate the position by meters. 
The GoToAnimator then continues to try to update the state to get near enough but it is nearly impossible because the rounding errors pertain and so it takes quite some time to get there. 

The first solution which is possible in the case of GoToAnimator is that we use directly Camera to set up the GoToAnimator instead of LookAt. I verified this solution and it seems to work at least to some extent verifying that the issue is due to the rounding errors.  The code for this change is in this branch https://github.com/NASAWorldWind/WebWorldWind/tree/feature/cameraGoToAnimator 

This issue won't be limited to the GoToAnimator only. It might happen anywhere where the LookAt interface is used to set up the state of the navigator. I am not sure what is the right path from this. Making sure that the arithmetic is precise is unviable as it would slow down the framework too much. In theory, we could keep the state as the LookAt as well as Camera, and update the other type by changes to one of them, but I believe that this will bring us whole other lot of issues such as synchronizing the state and I am not even sure it would fix this issue. 

I remember, that we already discussed this rounding issue before, but until now we didn't really run into visible glitches in the behavior of the globe. Internally we can make sure that we aren't using the LookAt interface unless necessary but this seems to me as an issue for external users, as it might lead the globe to behave in a bit strange way.

balhar-jakub:
GoToAnimator issue. The GoToAnimator works correctly now. The transformation to use Camera instead of transforming from and to LookAt all the time is correct notion. 

The issue with moving elsewhere than expected arose from the fact, that the user gives in position for LookAt which I was using as a resulting position in the Camera. The solution is to transform the final position so that the resulting camera position points to the right LookAt point.

pdavidc:
We're applying the work done in this pull request incrementally. Progress can be tracked in epic #279.

PR #102, opened 2017-03-31
fix horizon clipping of atmosphere

strangerintheq:
bug: using AtmosphereLayer with zoom and tilt will cause unnecessary clipping of image generated by SkyProgram

here is a bugfix

### DIFF
--- a/src/shaders/SkyProgram.js
+++ b/src/shaders/SkyProgram.js
@@ -138,6 +138,9 @@ define([
 
                         /* Transform the vertex point by the modelview-projection matrix */
                     '    gl_Position = mvpMatrix * vertexPoint;\n' +
+
+                        /* fix horizon clipping */
+                    '    gl_Position.z = gl_Position.w - 0.00001;\n' +
                     '}',
                 fragmentShaderSource =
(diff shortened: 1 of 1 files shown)


AkeluX:
@strikerM any thoughts on this?

strikerM:
In my opinion, we should use the same method as WorldWindAndroid for handling the clipping of the sky atmosphere.
[https://github.com/NASAWorldWind/WorldWindAndroid/blob/master/worldwind/src/main/java/gov/nasa/worldwind/WorldWindow.java#L1023](url)

I think the new Camera implementation should fix this issue.

@pdavidc @balhar-jakub

strangerintheq:
does this method can handle elevations?

strikerM:
I'm not sure I understand the question.
The Sky Atmosphere doesn't "follow" the terrain.

pdavidc:
@strikerM is right, the WorldWindAndroid method resolves this issue.

strangerintheq:
Clipping is cool, but what about negative elevations? Is method handle situation With max tilt and zoom in area in pacific ocean?

AkeluX:
Thank you for your inputs. We propose to continue the discussion in #105 as we won't take this pull request as is.

PR #118, opened 2017-05-10
Fixed clipping of Atmosphere 

emxsys:
Adjusted the far clip plane to account for the atmosphere and above-ground placemarks beyond horizon. 

_Before the fix, notice black line on the horizon:_
image

_After the fix, atmosphere appears correct:_
image

_Sunrise/sunset effect works now:_
image

_Placemarks now visible beyond the horizon:_
image

### DIFF
--- a/src/navigate/Navigator.js
+++ b/src/navigate/Navigator.js
@@ -111,9 +111,7 @@ define([
             // Compute the far clip distance based on the current eye altitude. This must be done after computing the
             // modelview matrix and before computing the near clip distance. The far clip distance depends on the
             // modelview matrix, and the near clip distance depends on the far clip distance.
-            this.farDistance = WWMath.horizonDistanceForGlobeRadius(globeRadius, eyePos.altitude);
-            if (this.farDistance < 1e3)
-                this.farDistance = 1e3;
+            this.farDistance = WWMath.horizonDistanceForGlobeRadius(globeRadius, eyePos.altitude) + 1.5e6;
 
             // Compute the near clip distance in order to achieve a desired depth resolution at the far clip distance.
             // This computed distance is limited such that it does not intersect the terrain when possible and is never


src/navigate/Navigator.js
```diff
@@ -111,9 +111,7 @@ define([
             // Compute the far clip distance based on the current eye altitude. This must be done after computing the
             // modelview matrix and before computing the near clip distance. The far clip distance depends on the
             // modelview matrix, and the near clip distance depends on the far clip distance.
-            this.farDistance = WWMath.horizonDistanceForGlobeRadius(globeRadius, eyePos.altitude);
-            if (this.farDistance < 1e3)
-                this.farDistance = 1e3;
+            this.farDistance = WWMath.horizonDistanceForGlobeRadius(globeRadius, eyePos.altitude) + 1.5e6;
```
AkeluX: @emxsys Instead of the fixed 1.5e6, WWA asks for the horizon distance of the atmosphere height. Would this be better here? See https://github.com/NASAWorldWind/WorldWindAndroid/blob/develop/worldwind/src/main/java/gov/nasa/worldwind/WorldWindow.java#L1041

pdavidc: The precomputed horizon distance in the original pull request is correct, but it has two issues: it's specific to Earth, and it's inconsistent with the pattern in Android. It seems to me that we could take this change if it were in line with what's done in Android. I suspect that @emxsys will agree that it'll be easier to keep the two SDKs synchronized if we avoid arbitrary differences.

pdavidc: If you have time to make the necessary changes, we can approve this PR. Otherwise, I'd be happy to open a new PR from an in-repository branch.

pdavidc:
Closed in favor of #172, which bring's Web WorldWind's far clip distance into parity with WorldWind Android.

PR #119, opened 2017-05-12
Add surface shape editor

AkeluX:


src/WorldWindow.js
```diff
@@ -979,6 +979,20 @@ define([
         };
 
         /**
+         * Indicates whether this world window contains a specified layer.
+         * @param {Layer} layer The layer to test.
+         * @returns {Boolean} true if this world window contains the specified layer, otherwise false.
+         */
+        WorldWindow.prototype.containsLayer = function (layer) {
```
AkeluX: Could we use indexOf instead? It might be useful to get the index of a particular layer in the array as well. There is something like that in LayerList in WWA at https://github.com/NASAWorldWind/WorldWindAndroid/blob/develop/worldwind/src/main/java/gov/nasa/worldwind/layer/LayerList.java#L66

src/geom/Vec3.js
```diff
@@ -523,6 +523,14 @@ define([
             return this;
         };
 
+        Vec3.fromLine = function(origin, t, direction)
```
AkeluX: Missing comment on the method.

src/shapes/ShapeAttributes.js
```diff
@@ -300,6 +300,25 @@ define([
             }
         });
 
+        ShapeAttributes.prototype.clone = function () {
```
AkeluX: The approach at other places is to use `new ShapeAttributes(baseShapeAttribute)` not a clone method. However, I see that Color has a clone method. We aren't consistent everywhere. Moreover, we might want to clone the color in the constructor of the attribute classes.

src/shapes/SurfaceCircle.js
```diff
@@ -74,6 +74,10 @@ define([
                 set: function(value) {
                     this.stateKeyInvalid = true;
                     this._radius = value;
+                    this.minorRadius = value;
+                    this.majorRadius = value;
+                    this.isPrepared = false;
```
AkeluX: All the isPrepared in the shapes don't seem to be used anywhere. Please remove if leftovers.

src/util/ControlPointMarker.js
```diff
@@ -0,0 +1,139 @@
+/*
```
AkeluX: I would put this class in shapes.

.idea/misc.xml
```diff
@@ -1,13 +0,0 @@
-<?xml version="1.0" encoding="UTF-8"?>
```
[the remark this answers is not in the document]
claudiaifrim: This file was added in .gitignore file in this commit https://github.com/NASAWorldWind/WebWorldWind/commit/123415de3ee436545b7973306b64b6ffbef748b5 by Miguel Del Castillo. This is the reason why this file is missing.

src/shapes/SurfaceCircle.js
claudiaifrim: isPrepared is used in SurfaceShape in prepareBoundaries method (see https://github.com/NASAWorldWind/WebWorldWind/blob/6d0fcb824b51817efe032a2b342f31c348d774fc/src/shapes/SurfaceShape.js#L532). For every change of any parameter (radius, heading, etc.) this method is called in order to recalculate boundaries.

.idea/misc.xml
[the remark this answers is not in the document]
AkeluX: In the meantime, @pdavidc reviewed the list of files to ignore from the IDE and this one is actually needed. See the latest version: https://github.com/NASAWorldWind/WebWorldWind/blob/develop/.gitignore

As this PR was started a long time ago, you might want to rebase your changes to the latest state of the develop branch.

AkeluX:
@claudiaifrim @pdavidc There is something really wrong with how GitHub is showing this pull request. If I attempt to create a new one, the change log is all clean and nice, but here it's like you have touch the entire framework. Do you know what's happening? Otherwise, we simply close this PR and open a new one.

claudiaifrim:
This is how is shown after the rebase. I can't say why. I think we should create a new pull request.

AkeluX:
Well, I've changed the base of the pull request to master and then back to develop. This has cleaned things up so I believe we can continue here. Interesting behaviour though.

I will review the most recent state.

PR #122, opened 2017-05-16
Use startsWith and endsWith from WWUtil in KML

AkeluX:


balhar-jakub:
I didn't want to do that change in this Pull request as in another one, there was change in the same file and I wanted to do this sequentially to prevent complexity in testing the thing twice and also possibly complicating one of the pull requests. 

Other than that it is ok and the code is updated in a fine way.

PR #155, opened 2017-06-30
Added CommonJS-like interface to almond wrapper.

### LINKED ISSUE #151: Modify library build process in order to expose a CommonJS interface
Previous work related to this was done in #139 in a mockup.

Beak-man:
- This changes the build artifact generation so it has a UMD interface.
   http://davidbcalhoun.com/2014/what-is-amd-commonjs-and-umd/

 - Build artifacts should now be capable of interfacing like Node modules
   ( with a bundler like Webpack or Browserify for
   frontend development), in addition to the previous interfaces
   (an AMD module or web browser object attached to the window object).

Closes #151

### DIFF
--- a/tools/wrap.start
+++ b/tools/wrap.start
@@ -1,5 +1,5 @@
 (function (root, factory) {
-    if (typeof exports === "object"){
+    if (typeof exports === "object") {
         // CommonJS-like
         module.exports = factory();
     } else if (typeof define === 'function' && define.amd) {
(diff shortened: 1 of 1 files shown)


tools/wrap.start
```diff
@@ -1,5 +1,8 @@
 (function (root, factory) {
-    if (typeof define === 'function' && define.amd) {
+    if (typeof exports === "object"){
```
pdavidc: @Beak-man there's a minor formatting irregularity at the end of this line. It should be:
   if (typeof exports === "object") {

Beak-man: Right, the space before the curly brace! It's fixed now. Thanks @pdavidc.

pdavidc:
Very simple. Looks great!

PR #161, opened 2017-07-12
Fixed wrong text in throwing error

GabrielePrestifilippo:
This message should be saying "elevationsForGrid" instead of "elevationsForSector"


pdavidc:
@GabrielePrestifilippo Thanks for catching these typos and for contributing a very straightforward fix.

PR #163, opened 2017-07-25
Add small app showing how to use ESA Sentinel data with the WMTS feature

AkeluX:


AkeluX:
Thank you for the review. I'm in touch with the data provider regarding the two issues you raised.

Issue #157, opened 2017-06-30
Test npm package with Webpack-based project


Beak-man:
Testing Web Worldwind as an npm package within  [React Starter Kit](https://github.com/facebookincubator/create-react-app) is producing errors that point out to Web Worldwind's 3rd party libraries. This boilerplate project utilizes Webpack.

Some 3rd party libraries inside src/util/ were probably adapted from Node modules. These were modified as AMD modules in order to operate with the rest of the codebase, which is AMD-based. It also seems that their Node-related CommonJS interfaces were deliberately blocked, as it can be seen [here](https://github.com/NASAWorldWind/WebWorldWind/blob/develop/src/util/jszip.js#L12) and [here](https://github.com/NASAWorldWind/WebWorldWind/blob/develop/src/util/proj4-src.js#L1). [Edit: After unminifying, it can be seen that this is not the case. They're UMD interfaces. AMD and CommonJS are both supported].

Before projects like Browserify or Webpack that provide web browsers with (formerly backend only) CommonJS interfaces became prevalent, and in the middle of the explosion of different module definitions and package managers in the JavaScript world, this was a clever and quick workaround to make these libraries work in an AMD environment with the rest of the project without rewriting their interfaces.

According to testing however, this seems to be at odds with WebWorldWind's usage as an npm package destined to be used in the frontend (Issue #136), since these libraries fail to load while working in a Node environment in the browser (e.g. under Webpack).

Ways to provide these libraries with more up-to-date interfaces will be investigated.

It bears to say that this is also related to issue #147. Instead of going back and forth between CommonJS and AMD module interfaces, the ideal solution would be to substitute these libraries with npm dependencies in the project. This will necessitate architectural refactoring, however, and it will be considered in the future.

Beak-man:
Some guidance:

https://github.com/webpack/webpack/issues/319

So maybe [these](https://github.com/NASAWorldWind/WebWorldWind/blob/develop/src/util/jszip.js#L12) [lines](https://github.com/NASAWorldWind/WebWorldWind/blob/develop/src/util/proj4-src.js#L1) are Browserify boilerplate?

The following project addressed something similar:

https://github.com/Matt-Esch/virtual-dom/pull/318/files?diff=split

Beak-man:
Files proj4-src.js and jszip.js will be provided with AMD interfaces so to get them on line with the rest of the project's modules.

Beak-man:
[...]

PR #169, opened 2017-08-04
Feature/npm

### LINKED ISSUE #171: Support scoping in npm package
As mentioned in testing task #159, the bundled library for npm had problems working within the  [scope of our npm organization](https://docs.npmjs.com/getting-started/scoped-packages). After some trial and error, it was found that while overriding the default filename entry point for the package was causing the issue. This override works without issue if the package is not scoped. This apparent behavior inconsistency was not found in npm's docs and it seems like an oversight.

The workaround was to copy the minified library artifact with a Grunt plugin and then rename it to the default entry point filename that npm uses ('index.js').

The inconsistency will be consulted with npm's support.

Beak-man:
These changes provide what's needed in order to produce a scoped npm package for Web WorldWind.

The work done relates to the following:

- Attributes were added to package.json in order to provide what's required to publish to npm, plus keywords that help into finding the package within npm's website. You may see how the package page will look like [here](https://www.npmjs.com/package/webww). An .npmignore file was added as well, that excludes everything from the project besides what we need bundled in the package.

- `proj4-src.js` UMD-like interface gave problems when the library was loaded into a CommonJS-like environment (like Node,  or the Webpack and Browserify defaults). Since we're compiling the build artifacts with an AMD-based tool anyway (RequireJS), the UMD interface was removed and the library was wrapped in a pure AMD interface that still works with RequireJS. This eliminated any issues while loading the library in a CommonJS environment.

- The aforementioned provided a functional npm package. Problems arose when it was tested as a package scoped within our npm organization. It was found that overriding the default script entry point filename was what was causing this problem, which is undocumented and probably an oversight from npm. A Grunt plugin to copy and rename files was added to our development dependencies in order to add an `index.js` artifact to bundle it within the package, besides our usual `worldwind.js` and `worldwind.min.js` artifacts. This enabled scoping to work.

With these changes, executing in the command line in the following (at the project's root):

`npm pack`

Should produce the following file:

`nasaworldwind-worldwind-0.0.1.tgz`

That can then be moved to whatever npm module-based project root folder in order to install it:

`npm install nasaworldwind-worldwind-0.0.1.tgz`

The package can be seen working in this simple [React test project](https://github.com/Beak-man/worldwind-react). The .tgz file it's included in the repository.

There are following considerations:

- Local image files are bundled, with the package, but they are not being used as of now. Web WorldWind looks for the image files in the root folder of the web document that calls the script. In the case of the example above, they had to be manually added into the '/public' folder, which is a requirement for that particular React boilerplate project, as [it can be seen here](https://github.com/facebookincubator/create-react-app/blob/master/packages/react-scripts/template/README.md#using-the-public-folder).

- `jszip.js` still has the same exact interface that `proj4-src,js` had, but it hasn't produced any bugs or loading issues. Testing the correct operation of this library will be done before the official release to npm's repository.

- The version number of the package will be updated for the official release.

Closes #171


.npmignore
```diff
@@ -0,0 +1,23 @@
...
+.npmignore
+.travis.yml
+config.json
+Gruntfile.js
+HowToBuildWebWW.md
+HowToCreateAndRunUnitTests.md
+karma.conf.js
+layout.tmpl
+package-lock.json
+package.json
```
pdavidc: Adding package.json and README.md to .npmignore is unnecessary.
https://docs.npmjs.com/misc/developers#keeping-files-out-of-your-package

Gruntfile.js
```diff
@@ -66,6 +73,7 @@ module.exports = function (grunt) {
     grunt.loadNpmTasks('grunt-contrib-requirejs');
     grunt.loadNpmTasks('grunt-jsdoc');
     grunt.loadNpmTasks('grunt-karma');
+    grunt.loadNpmTasks('grunt-contrib-copy');
```
pdavidc: Nice simple integration. How about moving grunt-contrib-copy next to grunt-contrib-requirejs to establish a clear pattern?

package.json
```diff
@@ -1,24 +1,25 @@
 {
```
pdavidc: The key changes look good, and some I'd like clarification on. What's the motivation behind the following changes:
- bugs
- directories
- repository url
- dependencies
- keywords

src/util/jszip.js
```diff
@@ -1,9155 +1,9412 @@
 /*!
```
pdavidc: Why have jszip.js and proj4-src.js changed?

.gitignore
```diff
@@ -20,4 +24,4 @@ standalonedata/
 
 worldwind.js
 
-worldwind.min.js
```
zglueck: I'm wondering about the order, would it be better to group index.js with the worldwind js files? Also to keep consistent with the other folders, should test-results get a trailing slash?

pdavidc: I had the same thought, but we've been ordering .gitignore and .npmignore alphabetically. Regarding the trailing slash, I have no preference provided we're consistent.

.npmignore
Beak-man: Thanks for the reference. I'll remove them. Do you think I should separate everything in .npmignore with newlines as in .gitignore?

pdavidc: Yes, that sounds best. It makes for a long file, but requires no thought when comparing it with other .ignore files.

Gruntfile.js
Beak-man: Will do!

package.json
Beak-man: - With bugs, I went with the string to clean it up and make it consistent with the rest of the attributes, as per [npm's documentation](https://docs.npmjs.com/files/package.json#bugs), since we're not adding a dedicated email for bugs, but it may be a good idea to consider it.

- Directories, repository url and dependencies were changed automatically, probably in one of the many 'npm init' tests I tried, and I didn't notice until now. I'll revert them as they were and make sure that's what's producing those changes just in case it's 'npm version', which also commits to git.

- I added the keywords manually since [they serve as search terms](https://docs.npmjs.com/files/package.json#keywords) to look for the package in npm's website. Suggestions on more of them are very welcome, of course.

.gitignore
Beak-man: Sure, sorry for the oversight again.

src/util/jszip.js
Beak-man: I went back and forth with them for so long that I had forgotten that I wasn't supposed to update them, as I did when I had the module resolution conflict to see if that solved it. That update may still be worthwhile but it's outside the scope of this pull request. I'll put the previous versions and add the pure AMD interface to proj4-src.

pdavidc: Thanks, Miguel. I thought proj4-src already has the pure AMD interface. Did we not handle that in a previous pull request?

.gitignore
pdavidc: No apology necessary. We haven't been clear about this convention, and I'm just now mentioning it.

package.json
pdavidc: Regarding bugs, I'm always in favor of using more standard forms. That sounds good.

Regarding directories, repository url and dependencies, let me know which command made the change. We can decide to keep or revert the changes depending on what's introducing them.

The keywords are a good idea. Please check with those used for the WebWW GitHub project to be certain that we're consistent.

Beak-man: Confirmed, it was npm init what added that 'git+' prefix and the empty dependencies object by default. And for some reason it also changes the directories and removes two of them. What's odd is that it leaves the rest intact. The documentation of [npm-init](https://docs.npmjs.com/cli/init) is not too good.

pdavidc: In that case, let's drop the changes introduced by npm init. Thank you for confirming.

Beak-man: I removed the changes induced by 'npm init', built the package and tested it again. It works OK.

src/util/jszip.js
Beak-man: To be honest I had the same feeling, but I double checked, and no, proj4-src.js and jszip.js have stayed the same in the develop branch since 2 years ago. Maybe we're thinking about #155, which added the CommonJS interface to the whole library, alongside its previously present AMD one.

I did mention proj4's pure AMD interface in [task #157](https://github.com/NASAWorldWind/WebWorldWind/issues/157#issuecomment-315753325), but that issue changed a lot while I was learning the implications of the related technologies. At first I was suspecting React, then Webpack, and it ended being something related to the CommonJS interfacing itself, development toolchain notwithstanding. After switching for the pure AMD interface in proj4, the package even loads inside Node (of course it halts when the library tries to access the web browser global object).

I'll revert to our old versions, add the pure AMD interface to proj4-src.js, and test the package.

pdavidc: That sounds great.

Beak-man: Tracing back my steps, I didn't leave the updated version of jszip, it was different because I left it unminified. Now it's exactly as it is currently in develop.

Beak-man:
@pdavidc Regarding proj4-src.js:

I went back and tried to wrap it with a pure AMD interface, but there's a concern.
I updated it because as it was originally, there didn't seem to we a way to apply the pure AMD pattern.

Originally, it had some sort of numbered nested CommonJS interfaces that RequireJS had no issue with, but precluded the library from being wrapped in a pure AMD interface.
````javascript
1:[function(_dereq_,module,exports){
//...
},{"mgrs":66}],2:[function(_dereq_,module,exports){
//...
},{"./deriveConstants":32,"./extend":33,"./parseCode":36,"./projections":38}],3:[function(_dereq_,module,exports){
//...
},{}],4:[function(_dereq_,module,exports){
//...
//etc
````
How these work exactly is beyond me, quite frankly. If you check line 1 of the original file, you can see that there's no clear place where to wrap the code with an AMD interface, and Node (and by extension Webpack) is not liking its overarching UMD implementation, even when I tried to eliminate the CommonJS section of the header. I assume it's because of the aforementioned nested interfaces. This is the first line of the old version, unminified:

````javascript
! function(e) {
    if ("object" == typeof exports) module.exports = e();
    else if ("function" == typeof define && define.amd) define(e);
    else {
        var f;
        "undefined" != typeof window ? f = window : "undefined" != typeof global ? f = global : "undefined" != typeof self && (f = self), f.proj4 = e()
    }
}(function() {
        var define, module, exports;
        return (function e(t, n, r) {
                function s(o, u) {
                    if (!n[o]) {
                        if (!t[o]) {
                            var a = typeof require == "function" && require;
                            if (!u && a) return a(o, !0);
                            if (i) return i(o, !0);
                            throw new Error("Cannot find module '" + o + "'")
                        }
                        var f = n[o] = {
                            exports: {}
                        };
                        t[o][0].call(f.exports, function(e) {
                            var n = t[o][1][e];
                            return s(n ? n : e)
                        }, f, f.exports, e, t, n, r)
                    }
                    return n[o].exports
                }
                var i = typeof require == "function" && require;
                for (var o = 0; o < r.length; o++) s(r[o]);
                return s
            })({
                    1: [function(_dereq_, module, exports) {
````

It finishes with the first nested interface. The library's authors did away with this pattern in the last version I built from their GitHub, and I was easily able to wrap it AMD-style.

That being said, I'm not sure what's this library used for, I just assumed it was for the transforms used when switching geographic projections. The library is switching projections OK. Do we need to test it further than that?

**Comment by pdavidc** [comment]
It's still unclear what modifications to proj4-src.js are in this pull request. In addition to the AMD style module, have you updated this library, or un-minified it?

Search for proj4-src in WebWorldWind using WebStorm. You'll see where it's in use and what needs to be tested.

File: package.json [hunk 6]
```diff
@@ -39,5 +43,15 @@
     "karma-junit-reporter": "^1.2.0",
     "karma-phantomjs-launcher": "^1.0.2",
     "karma-requirejs": "^1.1.0"
-  }
+  },
+  "keywords": [
```
pdavidc: Please synchronize these keywords with those in use on https://github.com/NASAWorldWind/WebWorldWind

package-lock.json
```diff
@@ -1,7 +1,8 @@
 {
-  "name": "web-world-wind",
-  "version": "0.0.1",
+  "name": "@nasaworldwind/webww",
+  "version": "0.1.2",
```
pdavidc: The project name and version in package-lock.json are incorrect. Is this file out of date?

.gitignore
```diff
@@ -6,12 +6,16 @@
 
 .idea/misc.xml
 
+.idea/preferred-vcs.xml
```
pdavidc: Are you certain this WebStorm project file should be ignored? What does it contain?

.npmignore
```diff
@@ -0,0 +1,41 @@
+.idea/
```
pdavidc: We're sorting .ignore lists alphabetically. This list is sorted by folder/file, then alphabetically.

package.json
Beak-man: OK, they're added now, and in lowercase. Do you think we should leave out "javascript"? It may be redundant as a search keyword in npm.

I also added "browser". Package authors tend to include that so that packages that are meant to be executed in a browser can be distinguished from the ones meant for NodeJS runtime:

https://www.npmjs.com/search?q=keywords:browser

package-lock.json
Beak-man: The new name corresponds to how the package will be identified in npm's repository, including its scoping within our org.

The version number was wrong with respect of what's in package.json, It's corrected now. 

I never manually changed package-lock.json, this was caused by ['npm version'](https://docs.npmjs.com/cli/version). When I found out that I didn't need to publish to npm to test the package (and hence, I needed to update the version number each time I published), I started to manually change the version in package.json, since 'npm version' commits to Git every time is executed and I didn't want to add noise to my commit history. I didn't review package-lock.json, however (the package works even with this discrepancy). Thanks for catching that.

Note that I left the nonsensical "0.0.1" version that is meant to be changed when we're ready to publish.

.gitignore
Beak-man: When I switched to working from the Mac to my Windows laptop, that file appeared. The Windows version of Webstorm created that file. In theory it should be harmless, and [JetBrains docs](https://intellij-support.jetbrains.com/hc/en-us/articles/206544839-How-to-manage-projects-under-Version-Control-Systems) don't say anything about it, but these are its contents:
````xml
<?xml version="1.0" encoding="UTF-8"?>
<project version="4">
  <component name="PreferredVcsStorage">
    <preferredVcsName>ApexVCS</preferredVcsName>
  </component>
</project>
````
It seems to relate to a [source control system](https://community.oracle.com/thread/4038126) of an [Oracle development environment](https://apex.oracle.com/en/) that I don't have installed. I don't remember seeing that file when we reviewed the contents of the .idea/ folder back when I was in Mac OS, so I chose to add it to .gitignore.

package.json
pdavidc: Nice catch with the "browser" keywords. Let's add "browser" to GitHub and keep the two lists identical. Imagine trying to decide what keywords to use where when the two are different. It's much more difficult to maintain.

package-lock.json
pdavidc: Thanks. So is package-lock.json correct now?

.gitignore
pdavidc: Here's the official JetBrains documentation on what to share:
https://intellij-support.jetbrains.com/hc/en-us/articles/206544839-How-to-manage-projects-under-Version-Control-Systems

Let's follow their recommendations, starting with WebWorldWind. Based on their recommendations we need to commit preferred-vcs.xml. That said, please be sure that the Windows laptop is running a current WebStorm version, and that the project is associated with GitHub (i.e. you're not using a command line client on the side).

Beak-man: I'll add the file then. I just confirmed that I have the latest version and I always use WebStorm's GitHub association while working on our project.

package.json
Beak-man: I think I cannot change it in GitHub. Can you?

Issue #148, opened 2017-06-27
Build process review: Library build improvement


Beak-man:
Modify and streamline (where applicable) the building process for the library. Besides having a positive impact on the project on its own, this will benefit integration with the Travis CI environment (i.e. faster builds), and will also set the stage for any modifications needed in the build process to expose the library as an npm package. Required for work in #136 

- [x] Review and continue work on branch feature/build_suggestions #143 
- [ ] Optimize requireJS task. It’s currently done twice. #146
- [x] Investigate whether external libraries in src/util can be refactored as npm dependencies #147

pdavidc:
Closing in favor of epic #537

Issue #147, opened 2017-06-27
Investigate whether external libraries in src/util can be refactored as npm dependencies


Beak-man:
Some of the libraries that WebWW utilizes may not exist in the npm registry, but it could be worthwhile to refactor the ones that do exist as npm packages to provide better dependency management.

Beak-man:
npm packages for every 3rd party library that is currently in src/util were found.

es6-promise -> https://www.npmjs.com/package/es6-promise
proj4-src -> https://www.npmjs.com/package/proj4
jszip -> https://www.npmjs.com/package/jszip
libtess -> https://www.npmjs.com/package/libtess

Beak-man:
Doing this would require changing our bundling mechanism, since RequireJS is not able to load CommonJS Node modules. Browserify can bundle projects with AMD and CommonJS modules with plugins, and Webpack is able to do all of them from the get go, even implementing ES6 modules. A newer option, SystemJS also exists.

A comparison between the four bundling tools can be found here: https://youtu.be/ahRNMEA4mWw

It could be worthwhile to consider an epic to change Web WorldWind's bundling methods. Besides providing us with dependency management for the 3rd party libraries that we require, it would also provide Web WorldWind with a 'standard' interface for extensions via npm packages.

pdavidc:
Very nice summary and explanatory video. Given your findings I agree with the decision to close this issue now, with the option consider an epic for changing the bundling mechanism in the future.

Beak-man:
New bundlers in the ever-changing JavaScript world:

https://parceljs.org/ - Supports CommonJS and ES6 module types.
https://rollupjs.org/ - Focused on bundling ES6 modules. Supports CommonJS and AMD modules through 3rd party plugins (https://www.npmjs.com/package/rollup-plugin-amd).

Guide on how to bundle ES6 modules alongside older definitions:
https://hackernoon.com/7-different-ways-to-use-es-modules-today-fc552254ebf4

**Summary:** 
Handling different JS module definitions is still quite sketchy and full of pitfalls. 

Support for the only official module type (ES6) is just partially being rolled out to major web browsers, thus new module bundlers seem to focus on enabling ES6 usage (basically becoming ES6 polyfills). That being said, support for CommonJS modules is still highly sought after, due to the prevalence of npm packages.

AMD modules are becoming really long in the tooth, and few people seem to care about ensuring supporting them in current projects.

PR #207, opened 2017-08-24
Feature/jszip AMD build

### LINKED ISSUE #202: Change bundling process for jszip
The third party library jszip, used for .kmz support in WebWW, has Browserify as bundling tool. When the library is built, Browserify links every module with some sort of nested CommonJS-like interfaces inside the file. [Webpack and Node seem to have issues with this linking method](https://github.com/webpack/webpack/issues/319). When Web WorldWind is consumed as an npm package, the problem with this kind of interfaces arise.

The same problem was happening with another 3rd party library, proj4-src. The authors of this latter library moved away from Browserify as bundling tool, so the issue with WebWW's npm package was avoided by updating the library and changing is interface to a pure AMD one, in order to get it in line with every other module of WebWW. But jszip is still being built with Browserify in its last version.

Changing the bundling method for jszip will be considered and investigated in order to bundle the library in a different manner (and hopefully, getting an AMD module as output). Webpack seems to be the ideal solution, SystemJS will be considered.

Beak-man:
### Description of the Change
Changed build of jszip from its Browserify-built UMD version (which is how the developer currently builds it) to a pure AMD version built by  Webpack.

The currently, Browserify-built version of jszip worked well up until we tried to bundle Web WorldWind within an npm package. This is related to an issue between how Browserify builds libaries and how implementations of the Node runtime [(e.g. Webpack or the Node command line runtime itself) parse them.](https://github.com/webpack/webpack/issues/319) This is similar to [the problems we encountered regarding the proj4-src library](https://github.com/NASAWorldWind/WebWorldWind/issues/157#issuecomment-313737728), which was built in a similar manner. While the proj4 authors changed its bundling method and we were able to solve issues related to it just by updating the library and wrapping it with a pure-AMD interface, jszip still uses Browserify and causes this issue. Problems with jszip were exposed when we tried to load the npm package via a shim, which was done to solve #171. Shimming most assuredly forced the Node runtime to parse the whole library before loading it, which didn't happen before implementing the shim file, which in its turn is why previously only the proj4 library caused problems within npm.

Since the problem with proj4 was solved by changing it to a pure AMD module, It was decided to try to bundle the library with Webpack instead, since this bundling is able to both consume AMD modules and produce an output this same type, in order to get the library in line with the rest of the AMD modules of Web WorldWind.

A procedure on how to obtain this version of the library will be described in #202. In summary, it can be built either [from the author's repository](https://github.com/Stuk/jszip), previous installation of Webpack needed, or it can be done [with my fork of jszip](https://github.com/Beak-man/jszip), which already has Webpack as development dependency and its required configuration. It's important to note that in order to maintain compatibility with the current jszip version in Web WorldWind, this AMD version was done with jszip's v2.x. branch, since the newer v3.x has some breaking changes, [as described here.](https://stuk.github.io/jszip/documentation/upgrade_guide.html)

Additionally, a comment regarding the library copyright and how it was obtained was manually added to the beginning of the library. If deemed necessary, I could add the whole procedure to obtain the AMD version of it in there.

### Benefits
The jszip library doesn't cause any more issues when Web WorldWind is compiled into an npm package. Not every use case of the library was tested, but .kml files are now able to be loaded inside a React project, as shown in #202. Previously the whole npm package was unable to load when Webpack's Node implementation arrived to the jszip section of Web WorldWind.

### Potential Drawbacks
This doesn't solve the core problem that we've had with Web WorldWind: We're still handling third party dependencies manually, and this modification of one of them is no exception. Our RequireJS, AMD-based library bundling has been at odds with npm, which is CommonJS-based. That been said, this exercise on changing bundling tools was useful into showing that it's relatively simple to switch from RequireJS to Webpack as a bundling tool, due to  Webpack's ability to consume many types of JavaScript modules in a single project (including AMD) right out of the box (unlike Browserify which is CommonJS-based and needs plugins and additional configurations to extend its module support). With Webpack, we wouldn't need to rewrite our current AMD modules, we could add CommonJS modules to the project as well, and even ES6 modules in the future.

The importance of the ability of interfacing with CommonJS modules is rooted in that Webpack would permit our project to handle its 3rd party dependencies as npm packages, which would enable the usage of npm for 3rd party dependency management, since it was found that every 3rd party library that we use is present in npm's repository (#147). I've yet to try [SystemJS](https://jspm.io/), which seems to have more advanced features than Webpack, but out of the three bundling tools tested up until now (RequireJS, Browserify and Webpack), I would strongly suggest to consider switching from RequireJS to Webpack in a future epic, which would relate to #148.

### Applicable Issues
Closes #202


Beak-man:
A guide on how to obtain the AMD version of the library was added to #202:

https://github.com/NASAWorldWind/WebWorldWind/issues/202#issuecomment-324704924

PR #218, opened 2017-09-15
[WIP] Add throttling of mousemove events in HighlightController

AkeluX:
### Description of the Change
This limits the amount of time-consuming picks triggered by the HighlightController, which are greatly affecting performance, in particular with a lot of shapes.

This solves an issue of the default HighlightController, which is in the core.

### Benefits
The default HighlightController as advertised in the examples and tutorials no longer significantly degrades the frame rate.

### Potential Drawbacks
Subsequent picks are slightly delayed and this can be noticed visually.

### Applicable Issues
None in GitHub Issues.

### DIFF
--- a/src/util/HighlightController.js
+++ b/src/util/HighlightController.js
@@ -20,9 +20,10 @@ define([
          * @constructor
          * @classdesc Monitors mouse-move and touch-device tap events and highlights shapes they identify.
          * @param {WorldWindow} worldWindow The WorldWindow to monitor for mouse-move and tap events.
+         * @param {Number} [fastestUpdateInterval=50] The fastest rate in milliseconds at which to update the highlighted items when the mouse moves.
          * @throws {ArgumentError} If the specified WorldWindow is null or undefined.
          */
-        var HighlightController = function (worldWindow) {
+        var HighlightController = function (worldWindow, fastestUpdateInterval) {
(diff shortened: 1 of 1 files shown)


AkeluX:
I will still make the threshold configurable and extract the throttle function into something reusable.

src/util/HighlightController.js
```diff
@@ -76,8 +76,30 @@ define([
...
+                        timeout = setTimeout(function () {
+                            last = now;
+                            func.apply(callContext, callArguments);
+                        }, remaining);
+                    } else {
+                        last = now;
+                        func.apply(callContext, callArguments);
+                    }
+                }
+            }
```
pdavidc: Missing semicolon.

AkeluX:
I recommend not to merge this for the moment. After thorough analysis, this doesn't bring enough benefits in my opinion. To be discussed next Tuesday.

pdavidc:
@AkeluX Shall we close this pull request while we discuss the best path forward for picking performance?

AkeluX:
@pdavidc Yes, let's close it. I propose to create an issue listing some of the leads we discussed on Tuesday.

pdavidc:
That sounds good. Would you be willing to open the new issue?

AkeluX:
Yes, will do.

PR #220, opened 2017-09-20
Tectonic Plates Layer should define properties on itself and not SurfacePolygon

balhar-jakub:
### Description of the Change

The Tectonic plates layer was incorrectly extending the SurfacePolygon when defining properties. Therefore when TectonicPlatesLayer has loaded the attribute property of SurfacePolygon is replaced by the one from SurfacePolygon. 

Otherwise, it breaks attribute usage on SurfacePolygon. 

### Applicable Issues

It is necessary for good work of shapes in Well Known Text
https://github.com/NASAWorldWind/WebWorldWind/pull/116


pdavidc:
The necessary changes to TectonicPlatesLayer have already been incorporated via #116

PR #232, opened 2017-10-10
Clone the color object when creating a new set of attributes from an existing one

claudiaifrim:
### Description of the Change
Fix a bug discovered in ShapeAttributes constructor. When this constructor is used it does not create a deep copy of the given object.

ShapeAttributes is a class used by many components.

### Benefits
You will be able to correctly create a deep copy of a given object. If you want to change a colour of an object placed on the globe, in this moment you will change the colour of the initial object also.

### Potential Drawbacks
None detected in this moment.

### Applicable Issues
If you want to change the colour of an object placed on the globe will be possible.


AkeluX:
This pull request solves the primary issue needed to be fixed and the other items discovered in the process were captured as #466.

PR #237, opened 2017-10-19
Feature/#227 file upload

### LINKED ISSUE #227: Create artifact upload script
Closes #224

Create the script which will upload the desired artifacts to the hosting service

### LINKED ISSUE #228: Integrate artifact upload script into continuous integration process
Closes #224

Add the upload script to the CI build and release process.

### LINKED ISSUE #244: Improve Artifactory publishing script input arguments
The `travis/publishToArtifactory.js` script is sensitive to argument file path formatting. Investigate potential solutions and alternative, more generic arguments.

zglueck:
Closes #227 
Closes #228 
Closes #244 

In addition to the merge of this code, travis will need the API key for the server instance.

### Description of the Change
Adds node script which deploys the contents of the npm tarball to the content server at Ames.

Provides an automated method of publishing release assets to our content server.

### Applicable Issues
#227, #228, #244


zglueck:
Additional investigation revealed Artifactory is a potential host for live examples as well. Including the `examples` and `src` folders and enabling a certain setting in Artifactory allows for the examples to properly display.

https://files.worldwind.arc.nasa.gov/generic-local/latest/examples/BasicExample.html

To add these directory to the upload would look like:
```javascript
deployDirectory(outputDir, version);
deployDirectory("examples", version);
deployDirectory("src", version);
deployDirectory(outputDir, "latest");
deployDirectory("examples", "latest");
deployDirectory("src", "latest");
```

zglueck:
Sorry about that, some of my testing closed this pr for some reason.

.travis.yml
```diff
@@ -35,6 +35,25 @@ deploy:
     on:
       tags: true
 
+  # Create the npm tarball, extract, and deploy files to Artifactory for development builds
+  - provider: script
+    script:
+      - source travis/createExtractTarball.sh
```
pdavidc: Ditch the source, favor ./ instead

pdavidc: The file name describes the steps executed in the script, but it is better to name it based on its semantic purpose, like prepareNpmPackage or something like that.

pdavidc: What’s the name “develop” doing here? That’s a branch name; we’re moving away from branch names in published assets.

travis/createExtractTarball.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash
+# Creates the npm tarball using the npm pack command then extracts the contents to a temporary directory
+
+npm pack
+# provides compatibility for both macOS and Linux
+npmtmpdir=`mktemp -d 2>/dev/null || mktemp -d -t 'npmtmpdir'`
+
+tar -xvf nasaworldwind-worldwind-*.tgz -C $npmtmpdir --strip-components 1
+
+export NPM_TMP_DIR=$npmtmpdir
```
pdavidc: This is too difficult to follow. It appears in the Travis file as an ambiguous variable, but it’s not till scanning this that I connected th dots. Could we extract to a predefined directory name in the project itself?

travis/deployToArtifactory.js
```diff
@@ -0,0 +1,145 @@
...
+ *  X-Checksum-Sha1: *, X-Checksum-Md5: *}}}
+ */
+var generateOptions = function (filename) {
+
+    var hash = calculateChecksums(filename);
+    // convert windows back slashes to forward slashes and change path to be relative
+    var normalizedFilename = filename.replace(inputDirectory, "").replace(/\\/g, "/");
+
+    var options = {
+        path: "/artifactory/generic-local/" + deploymentVersion + "/" + normalizedFilename,
```
pdavidc: Didn’t we settle on the repository “web”? Do we use one name for publishing and another for fetching?

.travis.yml
zglueck: I was using `source` to export the `NPM_TMP_DIR` environment variable, which the regular invocation doesn't do if I remember correctly. Unfortunately, the travis environment doesn't support `source` so I'm currently searching for an alternative.

I do like `prepareNpmPackage` although, I don't want to suggest this command is required for any npm publication actions.

`develop` was a placeholder for the web equivalent of `SNAPSHOT`

travis/createExtractTarball.sh
zglueck: With the difficulty in exporting the environment variable for the temporary directory, I believe the easiest path forward is exactly that, defining a directory name in the project.

travis/deployToArtifactory.js
zglueck: This is a typo, it should be web, testing was done with the generic-local repo which is how it snuck back in here.

travis/createExtractTarball.sh
pdavidc: That sounds reasonable. Provided the name is clear and there’s no collision, that sounds like a good plan.

.travis.yml
pdavidc: source makes sense for the reason you gave, but I gather that you’ll just drop it per your comment below on the temp for.

pdavidc: We can test publishing to SNAPSHOT now

zglueck: What about "unstable" as an alternative to SNAPSHOT?

SNAPSHOT is exactly what we are aiming for in Maven terms, but I haven't seen its use in any of the javascript libraries. I think it is confusing when the context's are mixed and unfamiliar to the broad javascript community. I'm also worried about confusion from parties with maven projects finding the SNAPSHOT version qualifier. Selecting a naming pattern that is more web-centric (lower case and not mixing any Maven or other unassociated build systems) will provide a more natural use case for web developers.

pdavidc: A web centric name sounds best, but we’re certainly not using unstable.

If there’s no clear choice, let’s drop it for this sprint.

travis/publishFolderToArtifactory.js
```diff
@@ -0,0 +1,138 @@
...
+    recursive(sourceDirectory, function (err, files) {
+        if (err) {
+            console.error(err);
+        }
+
+        for (var i = 0, len = files.length; i < len; i++) {
+            // remove the root directory path from the file path and convert to unix style slashes
+            var normalizedFilepath = path.normalize(files[i]).replace(path.normalize(sourceDirectory), "").replace(/\\/g, "/");
+            // format the path to conform to Artifactory path requirements
+            var normalizedPath = path.normalize("/" + destinationPath + "/" + normalizedFilepath);
```
pdavidc: How about path.join?

travis/publishFolderToArtifactory.js
```diff
@@ -0,0 +1,138 @@
...
+    var options = generateOptions(file, destinationPath);
+    var req = http.request(options, function (res) {
+        var chunks = [];
+
+        res.on("data", function (chunk) {
+            chunks.push(chunk);
+        });
+
+        res.on("end", function () {
+            if (res.statusCode !== 201) {
```
pdavidc: Don't we want res.statusCode >= 200 && res.statusCode < 400 ?

pdavidc: Is it possible that the HTTP status is something other than 201 if the file is being replaced?

zglueck: Permissions server side prevent replacement/override, this could be expanded to your suggested
`res.statusCode >= 200 && res.statusCode < 400` if you think it adds generality to the script.

zglueck: It looks like Artifactory returns 201 for replacement/override as well. Would it be better to expand the status code anyway?

zglueck: @pdavidc I think it would be best to stick with the 201 status code being the only valid response.

The object of the function, and the API function it is targeting, is to create/deploy an asset. The behavior of the API to this point has been to return a 201 on a successful deployment. I hesitate to accept a broader range of status codes as I'm not sure we understand the full implications of a different status code.

Issue #255, opened 2017-10-26
Remove inheritance between SurfaceEllipse and SurfaceCircle


Beak-man:
`SurfaceCircle` is currently implemented as a child class of `SurfaceEllipse`. Although this was intended to promote code reuse, it ended up in an odd class relationship where `SurfaceCircle` needs to set `SurfaceEllipse`'s `_minorRadius` and `_majorRadius` properties to update its state, besides its intended property `_radius`, exposing an overuse of inheritance in the case of these classes.

Making these two classes independent will improve `SurfaceCircle` code clarity.

Issue #241, opened 2017-10-20
Test SurfaceShapes not mentioned in #168


Beak-man:
Bug #168 mentions SurfaceCircle, SurfaceRectangle and SurfaceEllipse. Further testing suggests that other shapes are affected. (e.g. SurfaceSector, now included in the TestMovingSurfaceShapes example).

Closes #168

Beak-man:
SurfaceSector and SurfacePolygon seem to have similar issues. In their case they're also related to the zoom level.

Issue #240, opened 2017-10-20
Modify setters of SurfaceShapes


Beak-man:
After replicating the issue in #168 (see #236), modifying the setters in the SurfaceShapes class files (in the example, testing it with the 'heading' setter of both SurfaceEllipse and SurfaceRectangle), corrected the issue. Will modify the rest of the setters in these two objects, plus SurfaceCircle.

Closes #168

Issue #236, opened 2017-10-17
Exemplify SurfaceShape state updating bug


Beak-man:
Replicate bug mentioned in #168 in an example.

Closes #168

Beak-man:
Done in the example 'SurfaceShapeUpdating', comparing the updating of a placemark versus a SurfaceEllipse and SurfaceRectangle.

PR #260, opened 2017-10-27
Bug/#168 SurfaceShapes state updating propagation

### LINKED ISSUE #168: Various surface shape property changes do not fully propogate
`SurfaceCircle`, `SurfaceEllipse`, `SurfaceRectangle` cannot be updated after creation.

Firstly, `SurfaceEllipse` and `SurfaceRectangle` do not sufficiently dirty themselves when their property setters  are invoked: they set `this.stateKeyInvalid = true`, but should also set `this._boundaries = null` and `this.isPrepared = false` to allow their boundaries to be recomputed.

`SurfaceCircle` has this problem as well, but furthermore should also update its base ellipse's `this._majorRadius` and `this._minorRadius` when its `radius` property setter is invoked.

Beak-man:
### Description of the Change
As reported on #168, `SurfaceCircle`, `SurfaceEllipse` and `SurfaceRectangle` had issues while updating their state. An example was done that exposes this. At the same time, an investigation was done to review if other SurfaceShapes (`SurfacePolyline`, `SurfacePolygon` and `SurfaceSector`) had related issues. While testing, it was found that `SurfaceSector` was not updating either, as the class lacked setters to be able to mutate. This PR addresses these SurfaceShapes updating issues.

It's a necessity for Web WorldWind to have SurfaceShapes that are able to update their state. Most of these were already designed to do it, they only needed to properly set their dirty flags in order to do so.

### Benefits
- The state of the three SurfaceShapes mentioned in issue #168 can now be updated via a function on parent class `SurfaceShape`. It doesn't affect the mutability of `SurfacePolyline` and `SurfacePolygon`, which were updating OK before this fix.
- `SurfaceSector ` now has an `Object.defineProperties()` function and is able to receive a `Sector` to update its state.
- Changes the `isPrepared` flag of parent class `SurfaceShape` to the more descriptive name `boundariesArePrepared`.

### Potential Drawbacks
- `SurfaceCircle` needs to set properties of its parent class `SurfaceEllipse` for its state to update. Issue #255 was made to address this.
- It may be a good idea to add `this.stateKeyInvalid = true` to the new `resetBoundaries()` function of `SurfaceShape` to further simplify code in the children classes' setters (this was tested and it works).
- The included example `TestMovingSurfaceShapes` was intended at first only for bug hunting purposes and it grew quite long. It may be desirable to simplify it or remove it altogether before committing it to the repo.

### Applicable Issues
Closes #168, #236, #240, and #241.


src/shapes/SurfacePolyline.js
```diff
@@ -101,5 +101,10 @@ define([
             return SurfacePolyline.staticStateKey(this);
         };
 
+        // Internal. Polyline doesn't generate its own boundaries. See SurfaceShape.prototype.computeBoundaries.
+        SurfacePolyline.prototype.computeBoundaries = function(dc) {
+            return;
```
pdavidc: Same comment for the unnecessary return in SurfacePolygon.prototype.computeBoundaries.

src/shapes/SurfaceEllipse.js
```diff
@@ -85,6 +85,8 @@ define([
                 },
                 set: function(value) {
                     this.stateKeyInvalid = true;
+                    this.resetBoundaries();
+                    // this._boundaries = null;
```
pdavidc: What's the motivation for leaving this in? If it's no longer necessary, remove it. Commented out code leaves room for ambiguity in future development.

Beak-man: I intended to remove them, but I mistook while searching for them. Will do.

PR #261, opened 2017-10-27
Fix mismatching casing in name

MrDulka:
"OpenSearchNameSpaces" should be "OpenSearchNamespaces", matching the casing used everywhere else.

### DIFF
--- a/src/ogc/openSearch/descriptionDocument/DescriptionDocument.js
+++ b/src/ogc/openSearch/descriptionDocument/DescriptionDocument.js
@@ -7,17 +7,17 @@
  */
 
 define([
-        '../OpenSearchNameSpaces',
+        '../OpenSearchNamespaces',
         './OpenSearchUrl',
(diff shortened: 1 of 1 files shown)


MrDulka:
I am not sure if I understand that, could you please elaborate? It seems to me that it would be matching the spec now, is there any problem that I am not seeing?

pdavidc:
You're absolutely right @MrDulka. OpenSearchNameSpaces should indeed be OpenSearchNamespaces. I'd read your comment backward; my apologies for the confusion.

@AkeluX would you please take a look at this class name change? Are there any others that exhibit the same pattern?

AkeluX:
Thank you for raising this issue and offering a solution. We were actually still in the process of refactoring/cleaning this branch so good timing! Our latest version has been pushed this morning and already includes the fix proposed by this PR. I will thus close it. There are also other changes that were made for naming consistency. Our own PR to develop will follow shortly.

PR #262, opened 2017-10-29
Fix typo in withCredentials

MrDulka:
"witchCredentials" -> "withCredentials"

I think there shouldn't be any witches around :), but rather proper https://developer.mozilla.org/en-US/docs/Web/API/XMLHttpRequest/withCredentials

### DIFF
--- a/src/ogc/openSearch/OpenSearchRequest.js
+++ b/src/ogc/openSearch/OpenSearchRequest.js
@@ -33,7 +33,7 @@ define([],
             this.relation = options.relation || defaultOptions.relation;
 
             //A flag to indicate if the request should include credentials.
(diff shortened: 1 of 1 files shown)


pdavidc:
Thanks for catching another spelling/case typo in the new OpenSearch code, @MrDulka. We’ll refer this to @AkeluX and the ESA development team for integration.

This one’s especially timely given the upcoming Halloween holiday.

MrDulka:
Glad to help, this one really amused me. I just started to use the OpenSearch code and noticed these few small issues.

AkeluX:
That one was fun indeed! It was also already fixed on our side and is included in this morning's update, see 60aa0df and #261 for more information. Thanks again for your contribution.

PR #264, opened 2017-10-31
OpenSearch for EO implementation

strikerM:
### Description of the Change
OpenSearch for EO implementation for WebWorldWind

Allows WebWorldWind to work with OpenSearch for EO services.


examples/OpenSearch.js
```diff
@@ -0,0 +1,149 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+requirejs([
+        '../src/WorldWind',
```
AkeluX: Should load WorldWindShim instead.

src/ogc/openSearch/OpenSearchRequest.js
```diff
@@ -0,0 +1,106 @@
...
+    function (OpenSearchConstants) {
+
+        /**
+         * Constructs a request object with the specified options.
+         * @alias OpenSearchRequest
+         * @constructor
+         * @classdesc Object representing an OpenSearch request.
+         * @param {Object} options
+         */
+        var OpenSearchRequest = function (options) {
```
AkeluX: Please add comments on the public methods addHeader and getHeader. The properties of this class should also be documented as such.

src/ogc/openSearch/OpenSearchService.js
```diff
@@ -0,0 +1,271 @@
...
+         *  - discover used to get the description document;
+         *  - search used for querying the search engine.
+         *
+         * By default, this service can handle Atom for EO and GeoJSON responses.
+         * Other types of response formats can by added using the registerParser method.
+         *
+         * @param {String} url The URL of the OpenSearch description document.
+         */
+        var OpenSearchService = function (url) {
+            this._url = url;
```
AkeluX: The URL can be set here or as an option of the discover method. I think we should choose a single workflow, either a class is created per URL or the class is reusable and the URL given to the discover method. What do you think? Each option has further consequences in my opinion.

src/layer/OpenSearchLayer.js
```diff
@@ -0,0 +1,370 @@
...
+         * @param {Function} shapeConfigurationCallback The optional function called prior to creating a shape for
+         * the indicated GeoJSON geometry.
+         * This function can be used to assign attributes to newly created shapes.
+         * The callback function's first argument is the current geometry object.
+         * The second argument to the callback function is the object containing the properties read from the
+         * corresponding GeoJSON properties member, if any.
+         */
+        OpenSearchLayer.prototype.loadFromGeoJSON = function (geoJSONCollection, shapeConfigurationCallback) {
+            var shapeCb = shapeConfigurationCallback || this._shapeConfigurationCallback || null;
+            var polygonGeoJSON = new GeoJSONParser(JSON.stringify(geoJSONCollection));
```
AkeluX: We stringify the object here for having the parser convert it back right away. Please prepare a separate pull request adding objects as an option of the GeoJSONParser in addition to strings. Then, leverage this new option here.

AkeluX:
We need to add a reference to the specification in the document: 13-026r9

PR #267, opened 2017-11-01
Add option for configuring the resampling mode of surface images

### LINKED ISSUE #243: PNG rendered with blur effect
Hi all,
I have to show pixeled-PNGs on WWW but it seems that WWW applys a strange blur effect (as you can see [here](https://forum.worldwindcentral.com/filedata/fetch?photoid=157925)).

The image is rendered with the SurfaceImage class, as shown in the SurfaceImage example (http://worldwindserver.net/webworldwind/examples/SurfaceImage.html).

Here you can find:
- the png: https://cloud.services.meeo.it/f/e350597f4a294b5a95e7/
- the aux: https://cloud.services.meeo.it/f/10e32fb2fc834b84bc19/

I've already asked on the forum: https://forum.worldwindcentral.com/forum/web-world-wind/web-world-wind-help/157924-png-rendered-with-blur-effect

Could you help me?

Thanks, bye!
Sergio

AkeluX:
### Description of the Change
This added option allows drawing surface images without the blurry effect reported by some application developers.

This requires a change in the guts of texture management.

### Benefits
Increased flexibility in rendering options for the application developers.

### Potential Drawbacks
There is an additional check for every resource retrieved from the GPU resource cache manager, which might have an impact on performance.

### Applicable Issues
Closes #243


src/WorldWind.js
```diff
@@ -567,6 +567,18 @@ define([ // PLEASE KEEP ALL THIS IN ALPHABETICAL ORDER BY MODULE NAME (not direc
             FAILED: "failed",
 
             /**
+             * Indicates a linear filter.
+             * @constant
+             */
+            FILTER_LINEAR: "filter_linear",
```
AkeluX: The prefix FILTER was used in order to differentiate this constant from the other LINEAR constant in this class. In the first commit, I had actually created an enum class for holding these two, but moved to the standard approach afterwards. This might be a solution to consider for organising these constants in the future.

src/cache/GpuResourceCache.js
```diff
@@ -161,7 +161,13 @@ define([
             var entry = (key instanceof ImageSource)
                 ? this.entries.entryForKey(key.key) : this.entries.entryForKey(key);
 
-            return entry ? entry.resource : null;
+            var resource = entry ? entry.resource : null;
+
+            if (resource instanceof Texture) {
```
AkeluX: According to benchmarks, checking for the existence of the method is much faster than using instanceof, at least in FF and Chrome. This seems like a critical enough place to consider it. What do you think?

src/render/Texture.js
```diff
@@ -107,12 +136,29 @@ define([
...
         };
 
         /**
+         * Applies the configured texture parameters to the OpenGL context.
+         * @param {DrawContext} dc The current draw context.
+         */
+        Texture.prototype.applyTexParameters = function (dc) {
+            var gl = dc.currentGlContext;
+
+            // Configure the OpenGL texture magnification function. Use linear by default.
```
AkeluX: For the problem at hand, #243, changing only the magnification filter gives the best results in my opinion. Should we also consider the minification filter in this update? There is quite an added complexity to it due to the dependency on the logic introduced by the wrap mode in the constructor.

src/render/Texture.js
```diff
@@ -107,12 +127,43 @@ define([
...
+            gl.texParameteri(gl.TEXTURE_2D, gl.TEXTURE_MAG_FILTER, textureMagFilter);
+
+            // Try to enable the anisotropic texture filtering only if we have a linear magnification filter.
+            // This can't be enabled all the time because Windows seems to ignore the TEXTURE_MAG_FILTER parameter when
+            // this extension is enabled.
+            if (textureMagFilter === gl.LINEAR) {
+                // Setup 4x anisotropic texture filtering when this feature is available.
+                // https://www.khronos.org/registry/webgl/extensions/EXT_texture_filter_anisotrop
+                var ext = (
+                    gl.getExtension("EXT_texture_filter_anisotropic") ||
```
AkeluX: This used to be done once when creating the texture. Is this getExtension call expensive? Otherwise, we could still get it once in the constructor or even globally and simply use it here if present. What do you think?

src/WorldWind.js
```diff
@@ -578,6 +578,18 @@ define([ // PLEASE KEEP ALL THIS IN ALPHABETICAL ORDER BY MODULE NAME (not direc
              */
             FAILED: "failed",
 
+            /**
+             * Indicates a linear filter.
+             * @constant
+             */
+            FILTER_LINEAR: "filter_linear",
```
AkeluX: As discussed, should be LINEAR only.

src/WorldWind.js
```diff
@@ -578,6 +578,18 @@ define([ // PLEASE KEEP ALL THIS IN ALPHABETICAL ORDER BY MODULE NAME (not direc
...
+
+            /**
+             * Indicates a nearest neighbor filter.
+             * @constant
+             */
+            FILTER_NEAREST: "filter_nearest",
```
AkeluX: As discussed, should be NEAREST only.

pdavidc:
@AkeluX After our Skype call on Feb 20, I think this looks good as-is, pending the JSDoc correction in SurfaceImage.js.

AkeluX:
@pdavidc I've added the last changes I wanted to. Could you merge it?

PR #271, opened 2017-11-02
Enhanced logic for recalculating SurfaceShape extents

### LINKED ISSUE #270: Enhance SurfaceShape extent recalculation logic
Currently, SurfaceShape extents are not recalculated when underlying data such as the globe projection, vertical exaggeration, elevation model, etc. are changed. This leads to surface shapes behaving incorrectly and/or disappearing when these underlying pieces of data are changed. 

Logic similar to the what is used in AbstractShape utilizing shapeDataCache and currentData properties should be implemented for SurfaceShapes.

markpet49:
Closes #270 

### Description of the Change
Enhanced logic for recalculating SurfaceShape extents, takes into account projection changes, vertical exaggeration changes, etc.
This change prevents SurfaceShapes from behaving incorrectly and/or disappearing when changes are made to global attributes.
### Benefits
See above.
### Potential Drawbacks
There will be a slight performance penalty in some cases, but there were performance gains in others.
### Applicable Issues
#270


src/shapes/SurfaceShape.js
```diff
@@ -425,17 +437,61 @@ define([
...
+            shapeData.expiryTime = Date.now() + this.expirationInterval + 1e3 * Math.random();
+        };
+
+        // Internal. Intentionally not documented.
+        SurfaceShape.prototype.establishCurrentData = function (dc) {
+            this.currentData = this.shapeDataCache.entryForKey(dc.globeStateKey);
+            if (!this.currentData) {
+                this.currentData = this.createShapeDataObject();
+                this.resetExpiration(this.currentData);
+                this.currentData.verticalExaggeration = dc.verticalExaggeration;
```
pdavidc: Is the VE assignment here still needed?

markpet49: @pdavidc I suppose we can remove it which will just cause the currentData to immediately expire and get reset on the next render.

markpet49: The extent will be null anyway and so will cause computeExtent, etc. to fire.

markpet49:
@pdavidc Extraneous assignment has been removed.

pdavidc:
@markpet49 This PR looks good, but it needs to be brought up to date with develop before it can be merged.

markpet49:
@pdavidc The merge checked out OK.

PR #272, opened 2017-11-03
Enhancement/remove surface circle inheritance (#255)

### LINKED ISSUE #255: Remove inheritance between SurfaceEllipse and SurfaceCircle

Beak-man:
### Description of the Change
Refactored `SurfaceCircle` into an independent version of itself, instead of a child class of `SurfaceEllipse`.

The sole `radius:` setter of `SurfaceCircle` had to reference three different radii: `_majorRadius`, `_minorRadius` (private properties of `SurfaceEllipse`) _plus_ `SurfaceCircle`'s own `_radius`. This was needed in order for `SurfaceCircle` class to be mutable, but made the setter pretty confusing. It was decided that this was an overuse of inheritance, and that dissolving the class relationship between `SurfaceEllipse` and `SurfaceCircle` was preferable.

### Benefits
- Making SurfaceCircle an independent class removes the aforementioned reliance on more than one radii. 
- The boundary computing code for `SurfaceCircle` ended up being simpler than `SurfaceEllipse`'s, which seems fitting.
- Homogenizes the inheritance hierarchy to one level from `SurfaceShape` for all surface shapes. `SurfaceCircle` was the only one that was two levels beneath `SurfaceShape`.

### Potential Drawbacks
It's more code to deal with compared to a child class implementation. 

### Applicable Issues
Closes #255


markpet49:
Did you explore using one common utility method for SurfaceEllipse.computeBoundaries and SurfaceCircle.computeBoundaries?

Beak-man:
@markpet49, it was considered, but where to implement it if the aim was to break the inheritance relationship?

markpet49:
@Beak-man @pdavidc We could create a util/ShapeUtils.js class and put the common routine there. Having two sets of code for ellipse and circle creates the possibility of different results for what should be the same shape. Let's see if we can get some feedback from Dave.

pdavidc:
@markpet49 I was curious if a separate circle-specific implementation might be simple enough to warrant its own implementation. That's what @Beak-man has assembled here. While it is simple, your point is too compelling to ignore. We do introduce the maintenance headache of multiple circle/ellipse implementations.

I'm in favor of a single implementation in a utility class. Thinking ahead, will be other shapes that need to compute circle/ellipse boundary locations given a center location and a radius. The location for these common functions is less clear. /util is intended to hold general purpose classes that don't related to any one part of the system. Let's generate some ideas and see what we like.

Beak-man:
Maybe a /geom/Ellipse.js implementation? There's already a /geom/Rectangle.js class. Its functionality seems different from boundary computing, though.

pdavidc:
You're right that Ellipse, as we're currently envisioning it, would be functionally different from the other geom classes. That said, the geom folder is the right place, and there's likely a way we can design this capability such that is makes sense and establishes a good pattern.

@markpet49 and @Beak-man I think we ought to take this up as a separate iteration. We want to invest the necessary time to establish the right pattern. Are there any objections or other thoughts?

markpet49:
@pdavidc @Beak-man That sounds good to me.

Beak-man:
Good for me too.

PR #275, opened 2017-11-03
Refactor the examples reference method of the WorldWind library

### LINKED ISSUE #254: Refactor Examples to use new architecture
Closes #249 

Use the discussion and feedback from #253 to update all of the existing examples to the proposed architecture.

zglueck:
Closes #254

### Description of the Change
Adds a shim file to the examples. The shim provides a mechanism to switch the examples WorldWind module resolution from using the source files, to using the minified library. The shim for use of the minified library is not a part of this pull request.


examples/WorldWindShim.js
```diff
@@ -0,0 +1,9 @@
+/**
+ * This shim is used to switch between the individual files of WorldWind source and the minified library. It is
+ * not required for the use of WorldWind.
```
pdavidc: Good thinking adding a comment here.

@AkeluX talked this morning and would like to see a few points called out here:
- The WorldWind shim is not required to use WorldWind. You've done this already. How about we go one step further and suggest that applications include the minified library directly?
- The WorldWind shim is used to switch how we include WorldWind. Again, you've already done this. Let's mention the reason for switching: running examples from WorldWind source in the development environment vs running examples from the minified library online.

examples/SimplestExample.html
```diff
@@ -4,8 +4,8 @@
 <head lang="en">
     <meta charset="UTF-8">
     <title>WorldWind Example</title>
-    <!-- Include the Web WorldWind library. -->
-    <script src="http://worldwindserver.net/webworldwind/worldwind.min.js" type="text/javascript"></script>
+    <!-- Include the Web WorldWind library. You must first compile the library by running `npm run build` -->
+    <script src="../worldwind.min.js" type="text/javascript"></script>
```
pdavidc: Nice catch. This hard-coded path was well-intentioned (demonstrate how to use the hosted library) it has been a problem. That said, I'm not sure that worldwind.min.js will be placed in the parent directory in our deployment scenario. Could we use the shim?

Also, we can drop the additional comment; @AkeluX and I have plans to ensure this step isn't needed for folks exploring WorldWind.

zglueck: I believe the shim would require use of RequireJS, right? That change would then require all of the RequireJS boilerplate. Implementation wise no big deal, but this is one a few examples which demonstrates the library without RequireJS.

I know it is a bit presumptuous, but what about the actual (future) hosted library? https://files.worldwind.arc.nasa.gov/web/0.9.0/worldwind.min.js

The [Standalone](https://github.com/NASAWorldWind/WebWorldWind/blob/develop/examples/Standalone.html#L24) example also uses a relative path to the minified version. Do we need to change anything there?

pdavidc: Of course, the use of RequireJS is an oversight on my part. Let's deploy the examples in the same relative folder structure as appears in the development environment. In this case, worldwind.min.js is guaranteed to be in the parent folder.

examples/WorldWindShim.js
```diff
@@ -1,6 +1,8 @@
 /**
- * This shim is used to switch between the individual files of WorldWind source and the minified library. It is
- * not required for the use of WorldWind.
+ * This shim is used to switch between the individual WorldWind source files and the minified single file library for
+ * the WorldWind module. Switching allows locally developed examples to run from the individual WorldWind source files
+ * and hosted examples to use the faster to download minified library. A shim is not required for the use of WorldWind.
+ * The minified library (worldwind.min.js) is recommended for use in deployed applications.
```
pdavidc: Perfect

pdavidc:
@zglueck It appears that @Beak-man's change to a functional test example is in conflict. Would you please handle the merge conflict?

PR #282, opened 2017-11-08
Utilize time property of layer to determine sun location on Atmosphere Layer

### LINKED ISSUE #100: AtmosphereLayer should react to its time property
Every layer in WebWW has a time property. We are about to include an util for computing the position of the sun for the new star field layer. This new layer automatically adapts depending on the value of its time property. I thus propose that the atmosphere layer also react to its time property by computing the location of the light source automatically.

Nonetheless, I wouldn't remove the lightLocation property altogether because it might be useful to applications, which need to position the sun at a defined location. Whenever, the time is set, we could simply set the lightLocation accordingly. This would still give an application the opportunity to set the lightLocation manually.

What do you think?

zglueck:
Closes #100 

This pull request removes the `lightLocation` property and uses the layer time property (if not null) to determine the sun location in the Atmosphere layer.

### DIFF
--- a/src/layer/AtmosphereLayer.js
+++ b/src/layer/AtmosphereLayer.js
@@ -49,9 +49,6 @@ define([
             this._nightImageSource = nightImageSource ||
                 WorldWind.configuration.baseUrl + 'images/dnb_land_ocean_ice_2012.png';
 
-            //Documented in defineProperties below.
-            this._lightLocation = null;
-
             //Internal use only.
             //The light direction in cartesian space, computed form the lightLocation or defaults to the eyePoint.
             this._activeLightDirection = new Vec3(0, 0, 0);
--- a/src/layer/AtmosphereLayer.js
+++ b/src/layer/AtmosphereLayer.js
@@ -210,8 +210,8 @@ define([
 
             program.setScale(gl);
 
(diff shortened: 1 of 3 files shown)


src/layer/AtmosphereLayer.js
```diff
@@ -47,9 +49,6 @@ define([
             this._nightImageSource = nightImageSource ||
                 WorldWind.configuration.baseUrl + 'images/dnb_land_ocean_ice_2012.png';
 
-            //Documented in defineProperties below.
-            this._lightLocation = null;
-
             //Internal use only.
             //The light direction in cartesian space, computed form the lightLocation or defaults to the eyePoint.
```
pdavidc: Obsolete reference to lightLocation.

src/layer/AtmosphereLayer.js
```diff
@@ -226,7 +211,7 @@ define([
             program.setScale(gl);
 
             // Use this layer's night image when the light location is different than the eye location.
```
pdavidc: Looks like an obsolete comment; isn't the presence of time what matters now?

zglueck: That's right, missed the comment and the falsy value test right below, thanks! Updated in 20ff9ab

Issue #285, opened 2017-11-08
Publish development version of distribution folder


zglueck:
Closes #284 

Publish the assets in the distribution folder to a provider on every develop branch update.

PR #288, opened 2017-11-08
Publish dist folder contents to GitHub releases

### LINKED ISSUE #277: Publish distribution folder to GitHub releases
Closes #284

*This has been modified from the original intent.*

~~Add the deployment of the folder created in #276 to the continuous integration tasks.~~

Task #280 creates a distribution folder with all assets required to run the examples. Images, docs, license, and the readme are included in this distribution folder. This task will utilize the CI system to publish the distribution folder to GitHub releases.

zglueck:
### Description of the Change

This pull requests adds functionality to zip the `dist/` folder and then publish the zip file to GitHub releases using Travis. The configuration only allows publication on a tagged build.

### Benefits

Automates the deployment of the distribution assets to GitHub releases.

### Applicable Issues

Closes #277


pdavidc:
[DistributionArchive.patch.zip](https://github.com/NASAWorldWind/WebWorldWind/files/1455855/DistributionArchive.patch.zip)

zglueck:
@pdavidc compelling points, and thanks for the demonstration patch. I've been working through what you've provided and thinking about your comments. I look forward to discussing these topics more tomorrow.

> The distribution archive ought to sit alongside the distribution folder.

This does make sense; however, I thought one of the motivations to move to the `dist` folder was an attempt to prevent file pollution at the root? I agree the current arrangement could be prone to duplicate zipping of archives and needs to be changed. What if we move toward your early suggestion of a `build` folder? The current `dist` implementation could live inside (e.g. `build/dist`). The archive could sit alongside the folder without adding to the content in the root. Similarly, `test-results` could also live in `build`. Perhaps I'm over complicating the issue too :)

> The distribution archive needs the version in its filename.

Great change and idea. I pondered some kind of identifier embedded into the name as well. What about something more descriptive? The output of `git describe --tags` would provide a concise version string for the tagged release, while the added metadata for untagged commits would provide a unique, and traceable identification pattern. This could be held back on this pull request as we are targeting GitHub releases which require a tag, but something to keep in mind for #285.

> The distribution archive ought to be built separately.

I'm unsure if this convention is appropriate to enforce on javascript development, but I've noticed with Java build management systems that something like this wouldn't be possible. Both Maven and Gradle enforce the execution of previous build steps for tasks which depend on the output of previous steps (e.g. you can't test without first building, or you can't deploy without running the entire chain of build, test, jar). By not requiring the archive to be built following the fresh compilation and copying of the working directory I believe we're opening ourselves to potential deployment of a stale distribution archive.

zglueck:
@pdavidc I've implemented the changes we discussed this afternoon.

I checked the following use cases:
- Examples from the generated archive (both generated locally and from a deployment to GitHub releases)
- Use of the `npm pack` tarball in @Beak-man 's react demo
- Publication of the `assetsToPublish` to Artifactory
- Publishing to GitHub releases

If there are additional test cases, let me know, otherwise this pr is ready for review/merge.

PR #292, opened 2017-11-09
Fix Atmosphere initial state example

### LINKED ISSUE #291: Atmosphere example starts with day/night atmosphere
### Description

The work conducted to resolve #100 now shows the day/night transition on the initiation of the example instead of the pure Atmosphere display. The pure atmosphere display is still available if you cycle the "Sun Simulation"

### Steps to Reproduce

1. Open the Atmosphere example (notice there is a shadow and night)

Should show the atmosphere effect without the night portions

On initial load, the sun position is defined through the specification of Layer time property.

**Reproduces how often:** [What percentage of the time does it reproduce?]

Every time

OSX - Chrome

### Additional Information

I introduced this bug when I was working on #100 and noticed the effect today.

zglueck:
### Description of the Change

Remove the assignment of the current time to the `time` property of the layer thus preventing the determination of the sun position and display showing the day/night display instead of the pure atmosphere effect.

Returns the example to the original version.

### Applicable Issues

Closes #291


examples/Atmosphere.js
```diff
@@ -27,7 +27,6 @@ requirejs(['./WorldWindShim',
         
         var atmosphereLayer = new WorldWind.AtmosphereLayer();
         var timeStamp = Date.now();
```
pdavidc: The declaration and initialization of timeStamp seems out of place. Is there a better location?

zglueck: Yes, I think line 36 would be better. It would be grouped with the other variables declared for the simulation component.

PR #366, opened 2017-11-20
add shape editor feature

claudiaifrim:
### Description of the Change
Shape editor new feature.

It's a new feature that enable user to manage/move/rotate the shapes displayed on the globe.

### Potential Drawbacks
None.


AkeluX:
Split into smaller pull requests #454, #455, #456, #457 and upcoming

PR #371, opened 2017-11-20
Correct duplicate and misplaced copyright headers

pdavidc:
@markpet49 Would you please take a look at this?


src/formats/kml/geom/KmlLineString.js
```diff
@@ -1,3 +1,7 @@
+/*
```
markpet49: There are some code format issues in this file not related to this change. Can you do a "Reformat Code"?

Issue #302, opened 2017-11-14
WebGL texture warning in several examples


markpet49:
### Description

Several examples (Annotations, DeepPicking and others) display the following WebGL warning in the console:

[.Offscreen-For-WebGL-0x1633cf178600]RENDER WARNING: there is no texture bound to the unit 0

This warning occurs many times and appears to be linked to the mouse cursor scrolling over the canvas.

### Operating System and Version

Ubuntu 16.04/Chrome 61

balhar-jakub:
Confirm

### Operating System and Version
Windows 10 / Edge 40

Beak-man:
Confirmed in Windows 10/Chrome 62

Beak-man:
This warning appears in the following Examples and Apps:

Examples:
- Annotations
- Atmosphere
- Collada
- GeographicMeshes
- GoToLocation
- KmlExample
- Measurements
- MultiWindow
- Paths
- PickAllShapesInRegion
- PlacemarksAndPicking
- Shapefiles
- StarField
- TriangleMeshes (when hovering the pointer over the meshes)
- Wkt

Apps:
- SubSurface
- USGSSlabs
- USGSWells

All other examples/apps are exempt from this warning.
Tested in Chrome 62, Windows 10.

claudiaifrim:
Confirmed on Linux Mint, Chrome Version 62.0.3202.75 (Official Build) (64-bit)

pdavidc:
Chrome logs this warning when a WebGL draw call is made that meets two criteria:

1. The bound shader program expects a texture.
2. No texture is bound.

This warning appears in Chrome on all operating systems, and it appears whether or not the shader actually reads from the texture. This behavior is consistent with WebGL Security principles. Read the conclusion of the [WebGL Security white paper](https://www.khronos.org/webgl/security/) for details. @kenrussell, a WebGL developer and WorldWindJava collaborator alumnus, explains this in a conversation on the [three.js repository](https://github.com/mrdoob/three.js/commit/ae81860411fb397e5905b8282eb497351a61121b).

WebWorldWind drawing code needs to adjust to accommodate Chrome's warnings. We'll bind a small blank texture whenever the bound shader program *may* use a texture, even if we've disabled the texture read via a uniform variable.

PR #392, opened 2017-11-22
Repaired Text picking

### LINKED ISSUE #333: ScreenText example picking not working
### Description

In the ScreenText example, picking only works on the center text and not on the text in the corners. In addition, when the center text is picked the following error is displayed in the console:

WARNING: there is no texture bound to the unit 0

Ubuntu 16.04/Chrome 61

pdavidc:
- Both GeographicText and ScreenText picking failed when the value of BasicTextureProgram.opacity was anything other than 1.0.
- BasicTextureProgram is a shared WebGL shader program, which therefore shares its variable state with all drawing code that uses the program. Drawing code in one layer/shape can affect the drawing code of another when care is not taken to set all state appropriately for the current draw call.
- Modified Text to explicitly set BasicTextureProgram.opacity to 1.0 during picking.
- Consolidated unnecessarily repeated WebGL calls
- Collapsed unnecessarily nested control flow
- Grouped related WebGL calls
- Closes #333


markpet49:
@pdavidc On Ubuntu/Firefox: The fix works flawlessly. On Ubuntu/Chrome: Picking works flawlessly, but I still see the error message:
[.Offscreen-For-WebGL-0x3497eda66000]RENDER WARNING: there is no texture bound to the unit 0
I am satisfied with the fix, as is, if you are. I believe we have a backlog task to address this Chrome warning.

src/shapes/Text.js
```diff
@@ -492,10 +494,7 @@ define([
                 program.loadModelviewProjection(gl, Text.matrix);
 
                 // Set the pick color for picking or the color, opacity and texture if not picking.
```
AkeluX: This comment could be adapted to the change.

src/shapes/Text.js
```diff
@@ -468,6 +468,8 @@ define([
 
             if (dc.pickingMode) {
                 this.pickColor = dc.uniquePickColor();
+                program.loadColor(gl, this.pickColor);
+                program.loadOpacity(gl, 1);
             }
 
             // Compute the effective visibility. Use the current value if picking.
```
AkeluX: This isn't part of the fix, but there could be an `else` here instead of the second if. The comment is also detached from part of the logic, which is above.

pdavidc:
@markpet49 We do indeed have backlog task #302 that captures the WebGL texturing binding warnings.

@AkeluX My original take at this was to limit changes to an absolute minimum. That said, the stale comments and odd disjointed if/if statements have convinced me otherwise. I'll take the time to adjust the comments and the control flow to be more appropriate for the new logic.

pdavidc:
@markpet49 @AkeluX Please take a look at the latest version of this pull request. In light of @AkeluX's recommendations I've performed a light refactoring of Text's drawing code with the following intentions:
• Consolidate unnecessarily repeated WebGL calls
• Collapse unnecessarily nested control flow
• Group related WebGL calls

PR #401, opened 2017-11-28
Bug/324 wmts ie console error

### LINKED ISSUE #324: WMTS - The layer isn't loaded with error in console
### Description

Error in the console when loading: Object doesn't support property or method 'getAttribute'

Windows 10 / Edge 40

markpet49:
Closes #324 

### Description of the Change

Fix for issue #324 which prevents WMTS layers from displaying on Win7/IE11. Also included is an untested fix in WfsCapabilities.js. WfsCapabilities.js is currently not functional. 

### Potential Drawbacks

None known.


AkeluX:
We already check whether a node is an element or not in the Collada parser, but using the following:
```javascript
if (child.nodeType !== 1) {
    // Not an element
}
```
Although this might not be critical here, this is certainly faster than instanceof.

I've also seen the following in the framework:
```javascript
if (!xmlInput.getAttribute) {
    // Not an element
}
```
In my opinion, the first approach is clearer/cleaner.

In conclusion, we handle this in different ways within the framework. I'm not suggesting we unify this now. Nonetheless, we could agree on the way to check for this in general. What do you think of the nodeType approach?

Besides WFS, have you checked if other parsers are affected?

markpet49:
@AkeluX Thanks for the feedback. @pdavidc and I were chatting about this yesterday and observed that the node type was enforced in most areas as a side effect of checking the content of the localName attribute. We were thinking of removing 'instanceof' and checking for the correct localName. What do you think?

With respect to other parsers, I scanned the code for instances where we were checking for 'childNodes' and looked at the surrounding code. The bulk of these were protected by localName checks except for the instance I found in WfsCapabilities. I may have missed one or two in my visual scan, but testing each one would take a large amount of time.

markpet49:
@pdavidc The new version is ready for your review.

PR #403, opened 2017-11-28
Refactor TiledImageLayer fade out opacity value assignment

### LINKED ISSUE #353: Standalone example imagery tiles flicker
### Description

When tilted and near the surface, some imagery tiles flicker

### Steps to Reproduce

1. Load Standalone example and zoom into the surface
2. Tilt and Pan Around
3. Observe some tiles flicker

No Flickering

Flickering of some tiles

Android 8.0.0/Chrome 62/Nexus 5x

zglueck:
Closes #353 

### Description of the Change

This change moves the "reset" of a tiles opacity to the `fadeOutgoingTiles` method. The reset was previously conducted in the `addTile` method which preceded the fading stage. In some cases, the reset in the `addTile` method would override the opacity determined by the `fadeOutgoingTiles` method and cause the tile to "flicker" when the navigator was stationary.

It consolidates the opacity state change to the relevant method and eliminates the flickering effect observed in the Standalone example.

### Applicable Issues

#353

### DIFF
--- a/src/layer/TiledImageLayer.js
+++ b/src/layer/TiledImageLayer.js
@@ -304,18 +304,21 @@ define([
                 if (this.previousTiles.hasOwnProperty(tileImagePath)) {
                     tile = this.previousTiles[tileImagePath];
 
-                    if (tile.opacity > 0 && !current[tile.imagePath]) {
-                        // Compute the reduced.
-                        tile.opacity = Math.max(0, tile.opacity - visibilityDelta);
+                    // Determine the new fade out opacity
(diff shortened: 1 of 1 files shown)


zglueck:
@pdavidc and @markpet49 this is ready for review. Testing indicated resolution of #353 and no observed issues with other functionality.

zglueck:
The development of a more comprehensive test framework to validate this solution is necessary before it can be properly vetted. A temporary work around has been noted in the parent issue, #353.

PR #404, opened 2017-11-29
Refactored Text's drawing code to improve maintainabililty

pdavidc:
- Consolidated unnecessarily repeated WebGL calls
- Grouped related WebGL calls
- Split independent drawing steps into separate functions


pdavidc:
Thanks, @markpet49.

PR #443, opened 2018-01-05
Added text wrapping unit tests for TextRenderer

Beak-man:
### Description of the Change
Added a simple unit test for text wrapping functions of TextRenderer.

### Benefits
TextRenderer (previously TextSupport) had no unit testing until now.

### Potential Drawbacks
No substantial drawback. Maybe the addition of a Font is superfluous, but it makes the test less artificial.

### DIFF
--- a/test/render/TextRenderer.test.js
+++ b/test/render/TextRenderer.test.js
@@ -8,6 +8,7 @@ define([
     "use strict";
     describe("TextRenderer tests", function () {
 
+        // Mock TextRenderer.textSize() to avoid 2D context requirement.
         TextRenderer.prototype.textSize = function (text, font, outline) {
             return new Vec2(10, 50);
         };
--- a/test/render/TextRenderer.test.js
+++ b/test/render/TextRenderer.test.js
@@ -10,7 +10,7 @@ define([
 
         // Mock TextRenderer.textSize() to avoid 2D context requirement.
         TextRenderer.prototype.textSize = function (text, font, outline) {
-            return new Vec2(10, 50);
+            return new Vec2(text.length * 7, 16);
         };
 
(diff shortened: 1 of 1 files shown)


test/render/TextRenderer.test.js
```diff
@@ -0,0 +1,60 @@
...
+        it("Should throw an exception on missing text input", function () {
+            expect(function () {
+                var mockTextRenderer = new TextRenderer();
+                mockTextRenderer.wrap(null, 20, 100, myFont);
+            }).toThrow();
+        });
+
+        it("Should output '...' due to wrap height being less than textSize height", function () {
+            var mockTextRenderer = new TextRenderer();
+            var wrappedText = mockTextRenderer.wrap(testText, testText.length, 15, myFont);
```
markpet49: While not really a problem, using testText.length for width here and on line 36 is misleading because the width parameter is the number of pixels to allow. I would change this to a constant.

test/render/TextRenderer.test.js
```diff
@@ -0,0 +1,60 @@
...
+        it("Should output 'Lorem ipsum...' due to wrap width being less than textSize width", function () {
+            var mockTextRenderer = new TextRenderer();
+            var wrappedText = mockTextRenderer.wrap(testText, testText.length, 16, myFont);
+            expect(wrappedText).toEqual("Lorem ipsum...");
+        });
+
+        it("Should output every word on testText in different lines", function () {
+            var mockTextRenderer = new TextRenderer();
+            // Wrap line width less than textSize texture width
+            var wrappedLines = mockTextRenderer.wrapLine(testText, 9, null, myFont);
```
markpet49: You have an extra parameter on this call.

PR #444, opened 2018-01-08
Refactor NavigatorState

### LINKED ISSUE #416: Refactor NavigatorState
Closes #279 

Refactor the NavigatorState class by distributing the functionality to other classes.

markpet49:
Closes #416 

### Description of the Change

Removes the NavigatorState class by distributing the functionality to other classes.


markpet49:
@pdavidc Hi Dave. I think this PR is ready for another look.

pdavidc:
I like what you've done throughout the Web SDK here. It matches Android's advancements in all the ways that make sense, and deviates where it needs to. I have two questions:

• There are a few places where we're cloning a Matrix. It makes sense that we don't want to keep a reference, but this generates garbage. Could we replace these clones with a set/copy? 
• There are a few places where we're cloning a Line that appear unnecessary. What have I missed?

markpet49:
Good catch on the cloning, I will change those to set/copy. I will check the Line clones, but I know at least one of those instances translates the line in the underlying code, so a clone or set/copy is needed.

markpet49:
@pdavidc Hi Dave. I found the 2D panning issue (matrices not getting reset appropriately) and minimized cloning. Please have another look at this PR when you have a chance.

Issue #176, opened 2017-08-08
Refactor TextSupport as TextRenderer


pdavidc:
Closes #173 

Refactor and redesign TextSupport as TextRenderer to align WebWorldWind's text rendering components with that of WorldWindAndroid.

Align the Web TextSupport's class name, public properties, and public methods with Android TextRenderer.

Issue #173, opened 2017-08-08
Text outline control


pdavidc:
Applications need the capability to make text labels visible over the map background. Text outlines provide this capability and is usable with any combination of light/dark text and light/dark backgrounds. Supersedes #97 

Enhance text outline control by extending TextAttributes with the following properties:
- Outline enable bit
- Outline width
- Outline color

- [x] Refactor TextSupport as TextRenderer #176 
- [x] Consider decoupling text line wrapping functions into a new utility class #424 
- [x] Basic text outline attributes #178 
- [x] TextRender color support #174
- [x] Text outline color #175 
- [x] Reflect text attribute changes #177

PR #445, opened 2018-01-09
Refactor of TextRenderer properties and its createTexture function

Beak-man:
### Description of the Change
Modified TextRenderer to begin refactoring as an analogue of WorldWind Android's version. This change was focused on adding new properties to the class and modifying the createTexture function. Future refactors to other functions will follow.

TextRenderer unit test was also modified to reflect the changes in its constructor, and a new test was added.

This is a start on having Web WorldWind's and WorldWind Android's TextRenderer aligned between each other. It's also intended to be the first step on enabling text outlining color customization, as mentioned in issue #173.

### Benefits
Brings TextRenderer closer in terms of design between the Android and web version.

### Potential Drawbacks
- As it is now, every rendered piece of text has no outline by default (for Annotations, Placemarks, ScreenText, screen credits, etc). This could affect reading clarity depending on the imagery that's visualized in the globe, although the idea of course is to immediately revisit the outlining features after this PR.

- The TextRenderer requires a whole DrawContext to be fed to it. (This was also the case before this refactor), while only DrawContext.currentGlContext and DrawContext.pixelScale are used inside TextRenderer. Maybe limiting the consumption to this two properties instead of the whole object would be a better fit?

### Applicable Issues
Partially addresses #176

### DIFF
--- a/src/render/TextRenderer.js
+++ b/src/render/TextRenderer.js
@@ -113,7 +113,7 @@ define([
          * @param {String} text The text string.
          * @returns {Texture} A texture for the specified text string and font.
          */
-        TextRenderer.prototype.createTexture = function (text) {
+        TextRenderer.prototype.renderText = function (text) {
             var ctx2D = this.ctx2D,
                 canvas2D = this.canvas2D,
                 textSize = this.textSize(text, this.typeFace, this.enableOutline),
--- a/src/render/TextRenderer.js
+++ b/src/render/TextRenderer.js
@@ -114,7 +114,7 @@ define([
          * @returns {Texture} A texture for the specified text string.
          */
         TextRenderer.prototype.renderText = function (text) {
-            if (text !== null && text.length > 0) {
+            if (text && text.length > 0) {
                 var canvas2D = this.drawText(text);
                 return new Texture(this.dc.currentGlContext, canvas2D)
             } else {
--- a/src/render/TextRenderer.js
+++ b/src/render/TextRenderer.js
@@ -29,6 +29,7 @@ define([
     function (ArgumentError,
               BasicTextureProgram,
               Color,
+              Font,
               Logger,
               Matrix,
               Texture,
--- a/src/render/TextRenderer.js
+++ b/src/render/TextRenderer.js
@@ -20,6 +20,7 @@ define([
         '../error/ArgumentError',
         '../shaders/BasicTextureProgram',
         '../util/Color',
+        '../util/Font',
         '../util/Logger',
         '../geom/Matrix',
         '../render/Texture',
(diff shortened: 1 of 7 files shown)


src/render/TextRenderer.js
```diff
@@ -91,53 +109,49 @@ define([
...
+         * Creates a texture for a specified text string.
          * @param {String} text The text string.
-         * @param {Font} font The font to use.
-         * @param {Boolean} outline Indicates whether the text is drawn with a thin black outline.
          * @returns {Texture} A texture for the specified text string and font.
          */
-        TextRenderer.prototype.createTexture = function (dc, text, font, outline) {
-            var gl = dc.currentGlContext,
-                ctx2D = this.ctx2D,
+        TextRenderer.prototype.createTexture = function (text) {
```
markpet49: This should be refactored into two methods, TextRender.renderText(text) and TextRenderer.drawText(text) to conform to the Android model.

Beak-man: Right! I forgot about splitting up the function. Will do it ASAP.

Beak-man: @markpet49 Reconsidering splitting up the function, maybe it would be better to keep it a single one in WebWW.

`drawText()` in WWA is only used to assist `renderText()` in creating a bitmap, which is a concept that we don't apply in WebWW. Plus it's a `protected` function, and nothing else in the project calls it. As outlined in #176, the refactor focuses in public methods.

That being said, I should have renamed `createTexture()` in WebWW as `renderText()` in order to bring it up in line with WWA's version. They have the same signature and both are used by analogous classes in both projects.

What do you think?

src/shapes/Annotation.js
```diff
@@ -290,7 +290,7 @@ define([
             this.labelTexture = dc.gpuResourceCache.resourceForKey(labelKey);
 
             if (!this.labelTexture) {
-                this.labelTexture = dc.textRenderer.createTexture(dc, this.label, labelFont, false);
+                this.labelTexture = dc.textRenderer.renderText(this.label);
```
markpet49: One last thing, I think we should do as you suggested and default enableOutline to true. In Annotation, we should explicitly set and reset the outline property for now. e.g.
                    dc.textRenderer.enableOutline=false;
                    this.labelTexture = dc.textRenderer.renderText(this.label);
                    dc.textRenderer.enableOutline=true;
If we don't do this, the behaviour of the examples will change. As we discussed we'll revisit how this is done when we get to the attributes behaviour.

Beak-man: Done. The examples retained their behavior indeed.

Issue #279, opened 2017-11-06
Navigation interface enhancement


pdavidc:

pdavidc:
See pull request #101 for the initial full change set, and a description of issues encountered while implementing the initial changes.

PR #449, opened 2018-01-16
Minor LookAtNavigator cleanup

pdavidc:
- Moved LookAtNavigator properties used exclusively for gesture response from LookAtNavigator to BasicWorldWindowController.
- Adjusted unit tests to accommodate LookAtNavigator properties moved to BasicWorldWindowController.
- Relates to #279


pdavidc:
Thanks for catching that @markpet49 . I suppose Travis would have reported this overlooked unit test update as well, had I been more patient!

PR #454, opened 2018-01-28
Add a method for creating a Vec3 from a line

claudiaifrim:
### Description of the Change
Added new method 'fromLine' that returns the vector that corresponds with a specified line segment.
Will be used in Shape Editor.

Util method that could be used further in development.

### Benefits
Util method that could be used further in development.

### Potential Drawbacks
None.


AkeluX:
@claudiaifrim confirms that we can work with the functionality of the Line class instead.

PR #455, opened 2018-01-28
Align implementation of layer utility methods in WorldWindow

claudiaifrim:
### Description of the Change
Added new method 'indexOfLayer' that returns the index of a layer.
Will be used in Shape Editor.

Util method that could be used further in development.

### Benefits
Util method that could be used further in development.

### Potential Drawbacks
None.

### DIFF
--- a/src/WorldWindow.js
+++ b/src/WorldWindow.js
@@ -1251,6 +1251,20 @@ define([
             }
         };
 
+        /**
+         * Returns the index of a layer.
+         * @param {Layer} layer The layer to test.
+         * @returns {Number} Returns the index of the searched layer or -1 if it doesn't exists.
+         */
+        WorldWindow.prototype.indexOfLayer = function (layer) {
+            if (layer == null) {
+                throw new ArgumentError(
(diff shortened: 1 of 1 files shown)


src/WorldWindow.js
```diff
@@ -1251,6 +1251,20 @@ define([
             }
         };
 
+        /**
+         * Returns the index of a layer.
+         * @param {Layer} layer The layer to test.
+         * @returns {Number} Returns the index of the searched layer or -1 if it doesn't exists.
```
AkeluX: Check the grammar

src/WorldWindow.js
```diff
@@ -1251,6 +1251,20 @@ define([
...
+         */
+        WorldWindow.prototype.indexOfLayer = function (layer) {
+            if (layer == null) {
```
AkeluX: Verify whether this check is needed before indexOf

pdavidc:
The latest form of these changes look great. Thank you @AkeluX and @claudiaifrim.

PR #456, opened 2018-01-28
Add methods for getting a reference position and moving surface shapes

claudiaifrim:
### Description of the Change
Addition of getReferencePoint and moveTo to all Surface Shapes.

Addition of getReferencePoint and moveTo to all Surface Shapes, methods that will be used in Shape Editor feature.

### Benefits
Helper methods that can be used to change the location of the shape on the globe.

### Potential Drawbacks
None.


src/geom/Location.js
```diff
@@ -953,6 +953,40 @@ define([
...
+         * Computes a new set of locations translated from a specified location to a new location.
+         *
+         * @param {Globe} globe Globe used to compute translation.
+         * @param {Location} oldLocation The original reference location.
+         * @param {Location} newLocation The new reference location.
+         * @param {Location[]} locations The locations to translate.
+         *
+         * @return {Location[]} The translated locations.
+         */
+        Location.computeShiftedLocations = function(globe, oldLocation, newLocation, locations)
```
AkeluX: This is ported from Java as is, but wouldn't it better located in globe as a non-static method?

AkeluX:
@claudiaifrim as discussed, please clean up the broken rebase in this branch before we can have a look at it.

src/geom/Location.js
AkeluX: @claudiaifrim, we've discussed this with @pdavidc today. The logic here is more of the responsibility of globe, which knows about the projection, than location, which is simply a class representing a location (holding data). Globe already has this type of methods for other actions. We also see benefits in removing the double dependency this creates between Globe and Location. Consequently, could you move it as a non-static method of Globe?

src/shapes/SurfaceRectangle.js
```diff
@@ -192,5 +192,18 @@ define([
...
 
+        // Internal use only. Intentionally not documented.
+        SurfaceRectangle.prototype.getReferencePosition = function () {
+            return this.center;
+        };
+
+        // Internal use only. Intentionally not documented.
+        SurfaceRectangle.prototype.moveTo = function (globe, position) {
+            var locations = [];
+            locations.push(new Location(this.center.latitude, this.center.longitude));
```
AkeluX: Could the center property be used directly instead of creating a new location object?

src/shapes/SurfaceEllipse.js
```diff
@@ -212,6 +212,19 @@ define([
...
 
+        // Internal use only. Intentionally not documented.
+        SurfaceEllipse.prototype.getReferencePosition = function () {
+            return this.center;
+        };
+
+        // Internal use only. Intentionally not documented.
+        SurfaceEllipse.prototype.moveTo = function (globe, position) {
+            var locations = [];
+            locations.push(new Location(this.center.latitude, this.center.longitude));
```
AkeluX: Could the center property be used directly instead of creating a new location object?

src/shapes/SurfaceCircle.js
```diff
@@ -162,6 +162,19 @@ define(['../error/ArgumentError',
...
 
+        // Internal use only. Intentionally not documented.
+        SurfaceCircle.prototype.getReferencePosition = function () {
+            return this.center;
+        };
+
+        // Internal use only. Intentionally not documented.
+        SurfaceCircle.prototype.moveTo = function (globe, position) {
+            var locations = [];
+            locations.push(new Location(this.center.latitude, this.center.longitude));
```
AkeluX: Could the center property be used directly instead of creating a new location object?

src/shapes/SurfacePolygon.js
```diff
@@ -18,11 +18,13 @@
  */
 define([
         '../error/ArgumentError',
+        '../geom/Location',
```
AkeluX: It seems Location isn't used in this class anymore. Please remove this dependency if possible.

src/shapes/SurfacePolyline.js
```diff
@@ -18,11 +18,13 @@
  */
 define([
         '../error/ArgumentError',
+        '../geom/Location',
```
AkeluX: It seems Location isn't used in this class anymore. Please remove this dependency if possible.

src/globe/Globe.js
```diff
@@ -659,6 +659,38 @@ define([
...
+            var oldPoint = this.computePointFromLocation(oldLocation.latitude, oldLocation.longitude,
+                new Vec3(0, 0, 0));
+            var newPoint = this.computePointFromLocation(newLocation.latitude, newLocation.longitude,
+                new Vec3(0, 0, 0));
+            var delta = newPoint.subtract(oldPoint);
+
+            for(var i =0; i < locations.length; i++)
+            {
+                var point = this.computePointFromLocation(locations[i].latitude, locations[i].longitude,
+                    new Vec3(0, 0, 0));
```
AkeluX: This Vec3 could be declared before the for loop and reused to save some object allocations.

src/globe/Globe.js
```diff
@@ -659,6 +659,38 @@ define([
...
+                point = point.add(delta);
+                var newPos = new WorldWind.Position(0, 0, 0);
+                this.computePositionFromPoint(point[0], point[1], point[2], newPos);
+
+                newLocations.push(newPos);
+            }
+
+            return newLocations;
```
AkeluX: This actually returns an array of positions, not locations. That's what the Java implementation does as well, but I believe we should correct this here. @pdavidc what do you think?

pdavidc: Yes, I agree that returning locations would be best. In this context position's altitude is at best misleading, and at worst exposes the app to an error prone interpretation of the results. 

Regarding how this is accomplished, there's an important subtlety. I'd like to avoid adding a `computeLocationFromPoint` to Globe (on a related note, `computePointFromLocation` needs to be deprecated sometime in the future). These duplicate functions appear to be a nice convenience, but they have several problems: they hide necessary complexity (a point does not actually convert to a location and vice-versa), they complicate Globe's interface, and they expand the test suite needed to validate its Globe's correctness.

Given that, we'll need to use `computePositionFromPoint` as we do now, but convert each result to a location before its stored in newLocations. Memory overhead can be eliminated by declaring a scratch Position before the loop, in a manner identical to @AkeluX's comment about the Vec3.

src/shapes/SurfaceRectangle.js
```diff
@@ -192,5 +192,16 @@ define([
...
         };
 
+        // Internal use only. Intentionally not documented.
+        SurfaceRectangle.prototype.getReferencePosition = function () {
+            return this.center;
+        };
+
+        // Internal use only. Intentionally not documented.
+        SurfaceRectangle.prototype.moveTo = function (globe, position) {
+            this.center = globe.computeShiftedLocations(this.getReferencePosition(), position,
```
AkeluX: That's not what I had in mind. `this.center` is already a `Location` object. My questions was can we use it directly? Why are we creating a new `Location` object here? The same applies for `SurfaceEllipse` and `SurfaceCircle`.

src/shapes/SurfacePolygon.js
```diff
@@ -116,5 +116,39 @@ define([
...
+            else {
+                return null;
+            }
+        };
+
+        // Internal use only. Intentionally not documented.
+        SurfacePolygon.prototype.moveTo = function (globe, position) {
+            if(this.boundaries.length > 0 && this.boundaries[0].length > 2){
+                var boundaries = [];
+                for (var i = 0; i < this._boundaries.length; i++){
```
AkeluX: Here is the same issue in the for loop as mentioned for `computeShiftedLocations`. Please improve it as well with `for(var i = 0, len = this._boundaries.length; i < len; i++)`.

src/globe/Globe.js
AkeluX: @claudiaifrim please transform the returned list of `Position` objects into a list of `Location` objects as proposed by Dave here.

AkeluX:
@pdavidc We moved `computeShiftedLocations` from Location to Globe mainly to break the "package" tangling. I concur with your assessment that moving it into the hierarchy of shape classes would fit better the responsibility of the classes themselves as an additional benefit. As a starting point, we'll thus move it to `SurfaceShape`.

src/shapes/SurfaceShape.js
```diff
@@ -773,6 +775,40 @@ define([
...
+            var newPoint = globe.computePointFromLocation(newLocation.latitude, newLocation.longitude,
+                new Vec3(0, 0, 0));
+            var delta = newPoint.subtract(oldPoint);
+
+            for(var i = 0, len = locations.length; i < len; i++)
+            {
+                var point = globe.computePointFromLocation(locations[i].latitude, locations[i].longitude,
+                    result);
+                point = point.add(delta);
+                var newPos = new WorldWind.Position(0, 0, 0);
```
AkeluX: `newPos` can be declared before the loop and reused to store the positions returned by `computePositionFromPoint` without creating a new object each time.

pdavidc:
@AkeluX @claudiaifrim I like the way this came together. I think the time we took to integrate this functionality into WebWorldWind was worth it.

I have a question about SurfaceCircle's and SurfaceEllipse's implementations of moveTo: Isn't it enough to assign the center latitude and longitude to that of the incoming position?

claudiaifrim:
@pdavidc, yes, you are right - for SurfaceCircle and SurfaceEllipse we can just use the incoming position.
```
        SurfaceCircle.prototype.moveTo = function (position) {
            this.center = position;
        };
```

claudiaifrim:
@pdavidc, this pull request is already merged in develop. Should I include the commit with those changes in the pull request for Shape Editor (https://github.com/NASAWorldWind/WebWorldWind/pull/637)?

```

diff --git a/src/shapes/SurfaceCircle.js b/src/shapes/SurfaceCircle.js
index 0c09423..f9b66fd 100644
--- a/src/shapes/SurfaceCircle.js
+++ b/src/shapes/SurfaceCircle.js
@@ -168,8 +168,8 @@ define(['../error/ArgumentError',
         };
 
         // Internal use only. Intentionally not documented.
-        SurfaceCircle.prototype.moveTo = function (globe, position) {
-            this.center = this.computeShiftedLocations(globe, this.getReferencePosition(), position, [this.center])[0];
+        SurfaceCircle.prototype.moveTo = function (position) {
+            this.center = position;
         };
 
         /**
diff --git a/src/shapes/SurfaceEllipse.js b/src/shapes/SurfaceEllipse.js
index f145117..79600c8 100644
--- a/src/shapes/SurfaceEllipse.js
+++ b/src/shapes/SurfaceEllipse.js
@@ -218,8 +218,8 @@ define([
         };
 
         // Internal use only. Intentionally not documented.
-        SurfaceEllipse.prototype.moveTo = function (globe, position) {
-            this.center = this.computeShiftedLocations(globe, this.getReferencePosition(), position, [this.center])[0];
+        SurfaceEllipse.prototype.moveTo = function (position) {
+            this.center = position;
         };
 
         /**
diff --git a/src/util/ShapeEditor.js b/src/util/ShapeEditor.js
index 492feab..50652f5 100644
--- a/src/util/ShapeEditor.js
+++ b/src/util/ShapeEditor.js
@@ -677,7 +677,11 @@ define([
                 var p = new Position(0, 0, 0);
                 this.worldWindow.globe.computePositionFromPoint(intersection[0], intersection[1],
                     intersection[2], p);
-                this.shape.moveTo(this.worldWindow.globe, new WorldWind.Location(p.latitude, p.longitude));
+                if (this.shape instanceof SurfaceCircle || this.shape instanceof SurfaceEllipse) {
+                    this.shape.moveTo(new WorldWind.Location(p.latitude, p.longitude));
+                } else {
+                    this.shape.moveTo(this.worldWindow.globe, new WorldWind.Location(p.latitude, p.longitude));
+                }
             }
         };

```

PR #457, opened 2018-01-28
Add a ControlPointMarker shape that extends Placemark

claudiaifrim:
### Description of the Change
New class ControlPointMarker added - a visual marker with position and purpose (scope).

New class ControlPointMarker added - a visual marker with position and purpose (scope). This will be used in the implementation of Shape Editor.

### Benefits
Enables users to manage the geometry of a shape.

### Potential Drawbacks
None.

### DIFF
--- a/src/shapes/ControlPointMarker.js
+++ b/src/shapes/ControlPointMarker.js
@@ -1,0 +1,149 @@
+/*
+* Copyright 2015-2017 WorldWind Contributors
+*
+* Licensed under the Apache License, Version 2.0 (the "License");
+* you may not use this file except in compliance with the License.
+* You may obtain a copy of the License at
+*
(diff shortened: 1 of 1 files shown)


AkeluX:
@claudiaifrim This class doesn't bring a lot on top of the Placemark shape. What do you think of directly using Placemark without creating a specialised shape? The additional properties you need for the shape editor could be stored in the user properties of the renderable and the constants you define in this class could become constants internal to the shape editor.

https://github.com/NASAWorldWind/WebWorldWind/blob/74f1a8341cc48c984bf6d12802bff37406c32429/src/render/Renderable.js#L65

AkeluX:
@claudiaifrim confirms that we can work using only Placemark.

Issue #37, opened 2016-07-12
OpenStreetMap tiles from MapQuest broken


letmaik:
In the examples, e.g. http://worldwindserver.net/webworldwind/examples/PlacemarksAndPicking.html, when clicking on OpenStreetMap, it comes up with error tiles.

pdavidc:
MapQuest has terminated the free and open tile server that WebWW is currently using. We're looking into alternatives.

PR #458, opened 2018-01-29
Update tile with openstreetmap HOT tiles

rmarzocchi84:
Mapquest is not supported since 2016. Using this different tile the OSM layer works

**Note:** Filling out this template is required. Any pull request that does not include enough information to be reviewed in a timely manner may be closed at the
maintainer's discretion.

### Applicable Issues

### DIFF
--- a/src/layer/OpenStreetMapImageLayer.js
+++ b/src/layer/OpenStreetMapImageLayer.js
@@ -58,7 +58,7 @@ define([
             this.urlBuilder = {
                 urlForTile: function (tile, imageFormat) {
                     //var url = "https://a.tile.openstreetmap.org/" +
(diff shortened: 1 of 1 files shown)


PJHogan:
This involves replacing the old line from mapquest to instead access an openstreetmap server based in France. Though an interesting solution, OSM servers are typically not designed to be used in such a 'world-come-hither' way without permission. And World Wind would certainly become a 'heavy-hitter' on their systems.  https://operations.osmfoundation.org/policies/tiles/

pdavidc:
@rmarzocchi84 Thank you for your contribution. We have identified a potential MapQuest service is capable of supporting WorldWind community data loads. Feel free to follow issue #37 to track our progress.

PR #460, opened 2018-02-06
Fix WMTS TileMatrixSet selection

### LINKED ISSUE #113: WMTS subdivision selection
WMTS is not displaying correctly in WorldWind and an error is being thrown.

Service: https://basemap.nationalmap.gov/arcgis/rest/services/USGSShadedReliefOnly/MapServer/WMTS
Layer: USGSShadedReliefOnly
format: image/png

The error being thrown is "WmtsLayer.constructor: Tile subdivision not supported for layer : USGSShadedReliefOnly. Display until level 0"

WorldWind
ww_wmts

Cesium Reference of the same WMTS service
cesium_wmts

zglueck:
### Description of the Change

The WMTS specification is very flexible with the format and structure of provided imagery tiles. WebWorldWind currently only supports specific coordinate systems and tile subdivisions of the WMTS specification. The creation of a WMTS layer in WebWorldWind requires a configuration object with the TileMatrixSet specified. The TileMatrixSet includes the CRS and tile subdivision information.

The WmtsLayer object provides a utility function for creating the proper configuration object for the WmtsLayer constructor. The utility function checked for the compatible CRS; however, a check of the tile subdivision was left until the constructor. In some cases, this allowed a TileMatrixSet which included a compatible CRS but not a compatible tile subdivision to be specified in the configuration object. When the configuration object was provided to the constructor and the tile subdivision check occurred, an error is thrown and the WMTS layer is not displayed.

Notably, #113 demonstrates a case where two TileMatrixSet's are available from the server. Both have compatible CRS, but only the second has a compatible tile subdivision. The configuration object utility function ignored the subdivision incompatibility and provided a bad configuration to the WmtsLayer.

This change refactors the utility function and constructor slightly to conduct both the CRS and tile subdivision check. It is fully backwards compatible and ensures compatible CRS and tile subdivision checks happen in both the utility function and WmtsLayer constructor.

Corrects an oversight in the configuration of a WmtsLayer by checking the CRS and tile subdivision in both the config object creation function and the constructor.

### Benefits

More robust handling of available TileMatrixSet from an WMTS server.

### Applicable Issues

Closes #113

### DIFF
--- a/src/layer/WmtsLayer.js
+++ b/src/layer/WmtsLayer.js
@@ -481,7 +481,7 @@ define([
             // set negotiation.
             var supportedTileMatrixSets = wmtsLayerCapabilities.getLayerSupportedTileMatrixSets();
 
-            // Validate that the specified style identifier exists, or determine one if not specified.
+            // Validate that the specified TileMatrixSet exists and is compatible with WebWorldWind
             if (matrixSet) {
                 for (var i = 0, len = supportedTileMatrixSets.length; i < len; i++) {
                     if (supportedTileMatrixSets[i].identifier === matrixSet && WmtsLayer.isTileSubdivisionCompatible(supportedTileMatrixSets[i])) {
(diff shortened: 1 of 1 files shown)


pdavidc:
@zglueck The description of this change is concise and seems sound to me. That said, I'd appreciate a walkthrough of the code before we merge this.

Issue #174, opened 2017-08-08
TextRenderer color support


pdavidc:
Closes #173 

Provide the foundation for text outline control by enhancing TextRenderer to support interior and outline color.

TextRenderer's default state will be to draw text with a white interior and a black outline.

PR #462, opened 2018-02-07
Adressed bug on ScreenCreditController displaying text with outlines

Beak-man:
### Description of the Change
- Previous work on TextRenderer changed `ScreenCreditController` behavior. It was displaying the screen text with outlines, which shouldn't be the case. It was also using the `dc.TextRenderer.renderText()` function directly, instead of the new pattern used by Text and its subclasses of using `dc.renderText()`.

- Adds an example in `examples/ScreenCredits.html` for testing purposes.

### Benefits
This makes ScreenCreditController to behave as originally intended, and removes the last remaining usage `dc.TextRenderer.renderText()` outside of `DrawContext`.

### Potential Drawbacks
- Since `ScreenCreditControler` is not a `Text` sublcass, it has its own facilities to produce its text. It could be desirable to further modify it to better take advantage of its added `TextAttributes` member. Maybe even refactoring it as a subclass of `ScreenText` to avoid code repetition.

- As mentioned by @markpet49, we should probably avoid adding new examples indefinitely. The new example is intended for testing purposes of this PR and it may not be desirable to push it to develop.

### Applicable Issues
Part of #174


pdavidc:
@Beak-man The integration into ScreenCreditController is very concise. After reviewing the changes, I noticed that string credit textures are assumed to be grayscale. This is presently correct, and works as intended. However, this breaks from the new pattern established by Text, and could easily break. What if we associated text attributes with each string credit, rather than the controller itself? This could be accomplished in addStringCredit.

Regarding the example, I like what you're doing here. Let's omit this from the immediate pull request and touch base with @markpet49 about examples and functional tests later today.

Regarding potential follow-on changes to reduce duplicate code, I agree with your motivation, but the proposed design needs adjustment. ScreenCreditController is conceptually not a renderable, and as such it should never be a subclass of Text. Inheritance is often abused to reduce code repetition, leading to bloated and brittle class hierarchies. In cases like this, we look to composition rather than inheritance. So ScreenCreditController would compose itself with instances of ScreenText and ScreenImage, thereby leveraging their capabilities.

pdavidc:

PR #470, opened 2018-02-13
Apply layer opacity to surface shapes

### LINKED ISSUE #468: SurfaceShape layer opacity support
Apply the opacity value from the RenderableLayer to SurfaceShapes

Closes #134

zglueck:
### Description of the Change

This change applies layer opacity to SurfaceShapes.

### Applicable Issues

Closes #468

### DIFF
--- a/src/shapes/SurfaceShape.js
+++ b/src/shapes/SurfaceShape.js
@@ -947,43 +947,60 @@ define([
          */
         SurfaceShape.prototype.renderToTexture = function (dc, ctx2D, xScale, yScale, dx, dy) {
             var attributes = (this._highlighted ? (this._highlightAttributes || this._attributes) : this._attributes);
-            var drawInterior = (!this._isInteriorInhibited && attributes.drawInterior);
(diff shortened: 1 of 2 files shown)


src/shapes/SurfaceShape.js
```diff
@@ -947,43 +949,47 @@ define([
...
-            if (dc.pickingMode && !this.pickColor) {
-                this.pickColor = dc.uniquePickColor();
-            }
-
             if (dc.pickingMode) {
-                var pickColor = this.pickColor.toHexString();
+                if (!this.pickColor) {
+                    this.pickColor = dc.uniquePickColor();
+                }
+                ctx2D.fillStyle = this.pickColor.toHexString();
```
markpet49: @zglueck Although the default toHexString behaviour probably works fine here (I did not research it), I think it would be better to explicitly specify the value for the isUsingAlpha parameter. It seems more deterministic that way. It looks like the old code had this problem as well.

PR #472, opened 2018-02-14
Fix layer opacity application to renderables

### LINKED ISSUE #469: Consistent application of layer opacity
The application of layer opacity to renderables is often incorrectly handled resulting in alpha being applied twice.

Layer opacity is inconsistently applied during picking color assignment.

Closes #134

zglueck:
### Description of the Change

The application and use of layer opacity for regular and pick rendering differed throughout the SDK. This change unifies the application of layer opacity for picking and regular drawing and fixes some cases where opacity was applied twice.

### Applicable Issues

Closes #469

### DIFF
--- a/src/shapes/Annotation.js
+++ b/src/shapes/Annotation.js
@@ -451,7 +451,7 @@ define([
                 this.pickColor = dc.uniquePickColor();
             }
 
-            program.loadOpacity(gl, this.attributes.opacity);
+            program.loadOpacity(gl, this.attributes.opacity * dc.currentLayer.opacity);
 
             // Attributes have changed. We need to track this because the callout vbo data may
             // have changed if scaled or text wrapping changes callout dimensions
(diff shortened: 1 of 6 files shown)


pdavidc:
I have a couple of comments:
- The depthMask tests don't look right to me. The old conditional disabled the depth mask when opacity is less than one and pickMode is false. The new conditional has different semantics.
- It looks like Annotation's loadOpacity call is missing a test on pickMode.
- There are a few references to the old opacity variable that appear erroneous.

src/shapes/Annotation.js
```diff
@@ -451,7 +451,7 @@ define([
                 this.pickColor = dc.uniquePickColor();
             }
 
-            program.loadOpacity(gl, this.attributes.opacity);
+            program.loadOpacity(gl, this.attributes.opacity * dc.currentLayer.opacity);
```
markpet49: Would it be possible to deprecate AnnotationAttributes.opacity? It seems redundant to have an alpha channel on the colors and a separate opacity attribute.

zglueck:
Closing while fixing

zglueck:
@pdavidc I think I caught all of the references you were referring too.

PR #474, opened 2018-02-16
Update CSS Color String calculation and use

### LINKED ISSUE #471: Consistent CSS color use with Canvas 2D API
Refactor all fill/stroke CSS styles to use a common Color object to style string conversion.

zglueck:
### Description of the Change

This change corrects the calculation of CSS Color. It unifies the use of one string generation method (`toCSSColorString`) for all color rendering (including picking).

The method `toHexString` which also generates a CSS Color compatible string has been marked as deprecated. The CSS Color Module specification does not require hexadecimal color representations to contain an alpha channel. The new CSS Color compliant `toCSSColorString` incorporates an alpha channel to provide transparent colors.

### Applicable Issues

Closes #471

### DIFF
--- a/src/util/Color.js
+++ b/src/util/Color.js
@@ -339,7 +339,7 @@ define([
         };
 
         /**
-         * Create a rgba color string that conforms to CSS Module Level 3 specification.
+         * Create a rgba color string that conforms to CSS Color Module Level 3 specification.
          * @returns {string} A color string suitable for CSS.
          */
         Color.prototype.toCSSString = function () {
@@ -347,6 +347,7 @@ define([
                 green = Math.round(this.green * 255),
                 blue = Math.round(this.blue * 255);
 
+            // Per the CSS Color Module Level 3 specification, alpha is expressed as floating point value between 0 - 1
             return 'rgba(' + red + ' ,' + green + ' ,' + blue + ' ,' + this.alpha + ')';
         };
 
--- a/src/util/Color.js
+++ b/src/util/Color.js
@@ -348,7 +348,7 @@ define([
                 blue = Math.round(this.blue * 255);
 
             // Per the CSS Color Module Level 3 specification, alpha is expressed as floating point value between 0 - 1
-            return 'rgba(' + red + ' ,' + green + ' ,' + blue + ' ,' + this.alpha + ')';
+            return 'rgba(' + red + ', ' + green + ', ' + blue + ', ' + this.alpha + ')';
         };
 
         return Color;
--- a/src/util/Color.js
+++ b/src/util/Color.js
@@ -318,7 +318,7 @@ define([
          * because some uses reject a four-component color specification.
          * @param {Boolean} isUsingAlpha Enable the use of an alpha component.
          * @returns {string} A color string suitable for CSS.
-         * @deprecated since version 0.10.0
+         * @deprecated since version 0.10.0, use toCSSString for valid CSS color strings
          */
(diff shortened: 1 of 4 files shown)


pdavidc:
Simple and straightforward. The new method should be toCSSColorString, correct?

src/util/Color.js
```diff
@@ -338,13 +339,13 @@ define([
...
-        Color.prototype.toRGBAString = function () {
-            var red = Math.floor(this.red * 255),
-                green = Math.floor(this.green * 255),
-                blue = Math.floor(this.blue * 255);
+        Color.prototype.toCSSString = function () {
+            var red = Math.round(this.red * 255),
+                green = Math.round(this.green * 255),
+                blue = Math.round(this.blue * 255);
 
             return 'rgba(' + red + ' ,' + green + ' ,' + blue + ' ,' + this.alpha + ')';
```
markpet49: @zglueck Can we include a brief comment pointing out that the alpha channel is correctly inconsistent with the other channels?

zglueck:
@pdavidc my bad, I changed the name yesterday and I guess my brain decided to call it toCSSColorString in the pr. It is currently `toCSSString` in the library.

Which do you prefer?

src/util/Color.js
zglueck: no problem!

markpet49:
My vote is toCSSColorString it is easier on the eyes. In camel case, do acronyms maintain upper case? Maybe it should be toCssColorString?

zglueck:
Actually, most of the camel case acronyms I'm familar with in the library are lower case, e.g. WmsLayer.

I'll change to:
```
toCssColorString
```

markpet49:
Thanks Zach, I did a quick google on camel case acronyms and it would appear that the prevailing wisdom is to camel case any acronym longer than two characters. I certainly find it easier to read.

pdavidc:
The name toCssColorString looks best to me as well.

With WebWorldWind, we treated acronyms like any other word for the purpose of camel case names.

markpet49:
@zglueck Hey Zach. You're going to win the award for the most discussion for the fewest lines of code award. ;-)
Two things for you to consider:
1. Seeing as we went deep on the whole color calculation thing, could you add a small unit test to Color.test.js that ensures the calculation generates the expected result? 
2. Did you test colorFromByteArray with toHexString/toCssColorString to see if converting back and forth creates any drift in the color values? I'm not sure if we ever do that kind of round trip, but it would be nice if they were consistent.

zglueck:
@markpet49 I was going to make the same joke a minute ago :)

As for 1, I'll add a test and reopen the pr when it is ready.

For number 2, I actually overrode the SurfaceShape unique pick color assignment and tested subsequent picking results for an entire range of a color band. I setup 256 surface shapes and ensured the pick colors matched up correctly. I believe testing of just one band is sufficient as any floating point inaccuracies would be evident in the conversion of a single band.

zglueck:
Ok, added unit test.

markpet49:
@zglueck Thanks Zach. Two last things:
1. Color.js and Color.test.js need to have Code->Reformat Code run on them.
2. Add a short unit test that validates that toCssColorString and colorFromBytes generate similar results when you convert back and forth. We're trying to build up better unit test coverage and if we tweak the algorithm for colors, this would help us catch off subtle differences (off by one, etc).

zglueck:
@pdavidc and @markpet49 I've updated the unit tests and ensured the code has been formatted. I believe this one is ready for final review.

Issue #475, opened 2018-02-20
Screen credit hyperlink support


zglueck:
Closes #489

Add hyperlink capabilities to screen credit. This would allow a screen credit to be clicked and navigate to a configured site.

Consider placing the default capability of detecting screen credit picking in BasicWorldWindowController.

Depends on screen credit picking, which is provided by #467.

PR #476, opened 2018-02-20
Update Open Street Map data source

### LINKED ISSUE #37: OpenStreetMap tiles from MapQuest broken

zglueck:
### Description of the Change

The MapQuest provided OSM data was removed in 2016. This pr changes the data source to a WMTS source provided by @eox-a.

EOX IT Services requests attribution (which is included in this PR) as well as a hyperlink back to EOX (not currently implemented but requested in #475).

### Applicable Issues

Closes #37


zglueck:
Thanks for the feedback @markpet49 I’m going to do my best to address the points you’ve raised and continue the discussion on Open Street Map data in WorldWind.

Configurable URL
I believe it would be counterproductive to provide a configurable URL for the OpenStreetMap class as it is currently implemented in the feature/#37-osm-layer branch. Using EOX’s data means switching to the WMTS endpoint instead of the typical OSM slippy map. Providing the URL to another OSM slippy map would fail, as the current implementation relies completely on configuration from the WMTS capabilities document and the named layer “osm”. Providing a configurable URL and layer name just makes this another WMTS layer (see further discussion below).

Asynchronous Configuration in Constructor
Yep, it is a little ugly :) Perhaps those better versed in javascript inheritance can find a work around. I too prefer StarFieldLayer configuration/loading being kicked off from within the render loop; however, without “calling” the WmtsLayer constructor within the OSM layer constructor  the render loop (`doRender`) is never called during rendering. Is there a simple fix here I’m missing?

My question, what are we trying to accomplish with OpenStreetMapLayer? Should it be a convenience layer similar to BMNGLandsatLayer/BingAerialLayer, or a generic configuration supporting the OSM slippy map “format” without a data source specified?

markpet49:
@zglueck Regarding #1. Good point. #2 Perhaps rather than inheriting from WmtsLayer it could just use it like WmsTimeDimensionLayer does with WmsLayer?

markpet49:
At any rate, I thought it would be an opportune time to talk about our pattern for this sort of thing. I might be all wet. It wouldn't be the first time. ;-)

pdavidc:
@zglueck @markpet49 given the current OGC layer API in WorldWind, we are a bit limited. I'd like to refactor WorldWind's OGC layers/elevation models to address this concern, and would like to discuss that with you two at a convenient time.

@zglueck's position on #1 seems right to me.
@markpet49's question on #2 looks promising. How does this class look using composition rather than inheritance?

zglueck:
@markpet49  excellent catch on WmsTimeLayer approach. I'm updating the branch for composition instead of inheritance.

zglueck:
I've changed to a composition approach instead of inheritance. Let me know what you think.

pdavidc:
This follows the layer composition pattern established in WmsTimeDimensionedLayer very well. While the addition of another internal XmlHttpRequest is unfortunate, it is limited to one property and one method, and can therefore easily be refactored.

pdavidc:

PR #477, opened 2018-02-21
Change textBaseline to bottom for text rendering

### LINKED ISSUE #344: Firefox: Upper part of rendered text is cut off
### Description

Upper part of text is slightly cutoff.
The issue seems to affect any text rendered with WebWorldWind.

### Steps to Reproduce

1. Open the ScreenText example.

**Reproduces how often:** 100%

macOS High Sierra 10.13.1 / Firefox 57.0 (64-bit)

### Additional Information

ss_text

zglueck:
### Description of the Change

Canvas rendering Context 2D provides multiple options for defining the vertical placement of text when rendering text to a canvas. WebWorldWind currently specifies `top`; however, implementation differences by the browsers cause text to draw closer to the top in Firefox. This pr changes the definition of vertical placement to `bottom` which testing indicated was far more consistent across all major browsers (Chrome, Firefox, Safari, IE, Edge).

### Applicable Issues

Closes #344


pdavidc:
@zglueck How does this change the placement of text within the rendered image, if at all?

zglueck:
@pdavidc  The table below details the number pixels from the top of the image to the top of the "U".

Old | New | Browser
--- | --- | ---
0 | 4 | Firefox
3 | 5 | Chrome

And an image for reference:
comparison

The text is from the ScreenText example.

So we go from a delta of 3 pixels in terms of vertical placement to 1 for this particular case.

pdavidc:
Are you certain that the old vertical offset is still appropriate given a change in baseline? I'd imagine that this would need to be adjusted as well.

zglueck:
I don’t believe there is a compelling enough reason to adjust the vertical offset. The inconsistency between browsers definition of the anchor means we would have to optimize to a particular implementation or apply an initial offset based on the current browser.

Our calculation of a line size (based on text size, the outline, and a fraction of the text size) corresponds very closely to the Chrome/Safari top to bottom anchor point difference. By using the more consistent definition of the bottom anchor between Chrome/Safari/Firefox means our calculated line size is basically applying the same top to bottom anchor offset of Chrome to Firefox. As noted above, the vertical position difference between the different browsers is very similar and likely could only be corrected by an identification of the particular browser and custom offset being applied.

By using our existing vertical offset and the bottom anchor, the text render position in Firefox aligns very closely with Chrome/Safari. Additional testing by @Beakman on IE and Edge was not counter indicative.

PR #487, opened 2018-02-23
Move copyright symbol on OSM layer

zglueck:
@PJHogan I've moved the copyright symbol as requested. I think the site should remain EOX.at as www.EOX.at looks to be configured incorrectly.


PJHogan:
Yep, www.EOX.at needs an https:// to work correctly. Even though I knew that, I wasn't sure that ".at" would be enough to suggest the URL. Regardless, I concur with your sense of this.

PR #488, opened 2018-02-23
Add the EOX based OpenStreetMapImageLayer to BasicExample

pdavidc:


zglueck:
Do you want to disable Bing or leave it on?

pdavidc:
My goal here was to provide the option of viewing the EOX OpenStreetMap layer, but not change the example's defaults.

PR #493, opened 2018-02-27
Add GeoTiff parsing to ElevationModel and WcsEarthElevationModel

### LINKED ISSUE #483: WCS GeoTiff response support
Closes #482

Establish support for WCS responses in GeoTiff format. Connect the existing GeoTiff reader with the existing ElevationModel retrieval logic, making any necessary adjustments.

zglueck:
### Description of the Change

The `WcsEarthElevationModel` was not attached to `WorldWind` and the `ElevationModel` did not provide support for reading or parsing `image/tiff` responses from a WCS elevation request. This change adds `WcsEarthElevationModel` to `WorldWindow` and updates the content type handling to allow `ElevationModel` to read and parse GeoTiffs using the `GeoTiffReader`

The `GeoTiffReader` required a minor refactor and the addition of a method for reading and parsing data without an asynchronous network request.

### Applicable Issues

Closes #483


zglueck:
@markpet49 I believe there are two core code changes which have some applicability for testing.

1. The additional utility method of GeoTiffReader
2. The branching logic based on content type in ElevationModel

For number 1, designing a test ensuring the result from the new utility method (`readArrayBuffer`) is identical to the result from the utility method which incorporates its own network request (`readAsData`) would ensure consistency between the two operations. This test implicitly exercises the complete GeoTiff parsing functionality which introduces some ambiguity to the results, but at least we get some test coverage.

For number 2, mocking a result from a server request (simple xhr object this time) would demonstrate the correct branching. I would need to look into jasmine mocking techniques for mocking the GeoTiff creation to isolate tested functionality to the branching logic of the xhr provided retrieval format.

If there is value in the above tests, let me know. I'll pull this pr down, update, and submit for review once completed.

pdavidc:
I had a chance to discuss this with @AkeluX today. Here's what we came up with:
• We need to work within the existing public interface. readArrayBuffer has the effect of bolting on a secondary constructor.
• Let's adopt the pattern in GeoJsonParser, where the constructor accepts multiple types and the read methods operate on the type as appropriate. In the case of GeoTiff, those types are: string url; array buffer. 
• In the long term our format parsers need to be refactored as *pure* parsers. That is, they're not responsible for fetching the content, just parsing it. Having parsers do both is proving to be problematic.

zglueck:
I'm reopening with the changes discussed in the list serve.

zglueck:
@pdavidc and @markpet49 

Updates from the discussion and comments on Slack.

- The data and image retrieval methods have been renamed to getImageData and getImage
- The URL retrieval function name and callback format have changed to be more inline with Shapefile and GeoJSON
- Accessing a particular image from a multimage GeoTiff requires additional work to the API and was not implemented

src/formats/geotiff/GeoTiffReader.js
```diff
@@ -53,45 +53,32 @@ define([
...
+         * @param {String} ArrayBuffer The ArrayBuffer of the GeoTiff
          * @throws {ArgumentError} If the specified URL is null or undefined.
          */
-        var GeoTiffReader = function (url) {
-            if (!url) {
+        var GeoTiffReader = function (arrayBuffer) {
+            if (!arrayBuffer) {
                 throw new ArgumentError(
-                    Logger.logMessage(Logger.LEVEL_SEVERE, "GeoTiffReader", "constructor", "missingUrl"));
+                    Logger.logMessage(Logger.LEVEL_SEVERE, "GeoTiffReader", "constructor", "missing ArrayBuffer"));
```
markpet49: My understanding is that "missingArrayBuffer" should be one word and match a key in the messageTable in Logger.js. I assume this is done for translation purposes, I think we're missing a lot of keys.

src/formats/geotiff/GeoTiffReader.js
```diff
@@ -207,6 +218,25 @@ define([
...
+            return this.createImage();
+        };
+
+        /**
+         * Creates a typed array based on the contents of the GeoTiff.
+         *
+         * @return {TypedArray}
+         */
+        GeoTiffReader.prototype.getImageData = function () {
+            return this.createTypedElevationArray();
```
markpet49: @zglueck I'm curious as to why we refer to Elevation in this class, do you know? It makes it sound like we're making an assumption about the data content. Presumably that routine predates your changes, so maybe that's a future PR.....

src/formats/geotiff/GeoTiffReader.js
pdavidc: Our early convention was to create keys for common messages, and put one-off messages inline. So both patterns are valid.

The arrayBuffer argument name is likely to come up again, so I think Mark's recommendation should be implemented.

src/formats/geotiff/GeoTiffReader.js
pdavidc: GeoTiffReader's initial interface was constructed to very literally satisfy the use cases we had in mind: load an image as a Canvas, load an image as elevations.

That said, the naming doesn't isn't reflective of anything about GeoTiffReader or Tiff files. GeoTiffReader is ripe for a refactoring.

pdavidc:
@zglueck Please take a moment to consider @markpet49's questions above. Aside from those, there's one thing missing which @markpet49 mentioned in his initial comments: unit tests. What's involved with establishing Jasmine unit tests that exercise at least one RGB tiff and one WCS response tiff?

zglueck:
@markpet49 and @pdavidc thanks for the review and comments, I've added some basic unit tests to ensure we're getting expected response from the API. The are sourced from our GeoTiff example (for RGB) and a WW26 WCS request. Both have been resized using gdal to reduce bloat to the repository. Values tested were verified using gdal as well.

src/formats/geotiff/GeoTiffReader.js
markpet49: @zglueck Hi Zach, I think the vote was to go with missingArrayBuffer (no space) here.

zglueck:
@markpet49 I've added the logging message to the message table, I missed your note.

As for the flavors of Tiff, the GeoTiff reader does throw an exception if the tiff is not of the GeoTiff variety. I'm not as familiar with the complete parsing chain, but I know it supports multiband, subfiles, tiling, and pack bits compression. I haven't tested with a tiled tiff, but during testing with Android, I noticed WW26 only produces striped Tiffs, and the default for GeoServer was stripped as well (no compression for either).

Log warning messages are generated for incompatible compression formats, should these be exceptions instead?

markpet49:
@zglueck Hi Zach. I tried these four variations with the RGB unit test:

markpet49:
gdalwarp -co "TILED=YES" -co "BIGTIFF=YES" -t_srs "+proj=longlat +datum=WGS84" -dstalpha -overwrite rgb-test-orig.tif rgb-test.tif
gdalwarp -co "TILED=YES" -t_srs "+proj=longlat +datum=WGS84" -dstalpha -overwrite rgb-test-orig.tif rgb-test.tif
gdalwarp -co "TILED=YES" -t_srs "+proj=longlat +datum=WGS84" -overwrite rgb-test-orig.tif rgb-test.tif
gdalwarp -co "COMPRESS=NONE" -t_srs "+proj=longlat +datum=WGS84" -overwrite rgb-test-orig.tif rgb-test.tif

The first one failed the type check and, as I recall, the other three simply returned bad data. I didn't delve into the particulars, perhaps you can take a look and see what's happening. Thanks.

markpet49:
@zglueck I looked through the TIFF code some more and it seems likely that these variations were not supported in prior versions of the code either, so once you investigate, we can discuss whether or not to address better TIFF support in another issue.

markpet49:
Here is an example of what I did in the last TIFF parser I wrote when the code didn't support certain variations. 
case TAG_TILE_WIDTH:
    tileWidth=(Integer) getValue(tifFile,Count,Offset,FieldType);
    break;
case TAG_COLOR_MAP:
    colorMap=(int[]) getValue(tifFile,Count,Offset,FieldType);
break;
.
.
.
// only works for uncompressed RGB TIFs
if (tileWidth>0 || colorMap!=null)
    throw new Exception("clipImage: unsupported TIF format");

AkeluX:
@markpet49 @zglueck here are a bit of history and a couple of references. As you've noticed, the original implementation didn't foresee supporting all possible variations of GeoTIFF. At the time, we looked at the images returned by a pool of WCS servers to determine the options to implement first. The parser currently supports tile- and strip-based files with some photometric interpretations (https://github.com/NASAWorldWind/WebWorldWind/blob/fae3aa0a488598c210f8f90d49990bc7d072eb7e/src/formats/geotiff/GeoTiffReader.js#L348) and a limited set of compressions (https://github.com/NASAWorldWind/WebWorldWind/blob/fae3aa0a488598c210f8f90d49990bc7d072eb7e/src/formats/geotiff/GeoTiffReader.js#L534). Unsupported variations raise a warning message in the console as pointed out by @zglueck. This has been enough for us so far. I hope this helps in your current assessment.

zglueck:
@markpet49 variations 2, 3, and 4 do parse and display correctly when used in the GeoTiffExample. The warp changed the dimensions of the images. As the unit tests are expecting a certain size and value at a specific index, the test tifs failed. Adding `-ts 100 100` to maintain the original image dimensions made the unit tests work in my environment with your transforms.

Should the unit tests be modified to provide a more generic indexing operation and not check the image size?

markpet49:
@zglueck Thanks for looking in to this. I didn't notice that gdalwarp changed the dimensions, that makes sense. I think your unit tests are fine as is.

markpet49:
I have no further comments, but, I will leave this open for feedback from @pdavidc .

PR #496, opened 2018-03-01
Repaired multi-window user input navigation

### LINKED ISSUE #495: Navigation gestures in multi-window app affects all globes
Mouse and touch input in a multi-window app erroneously affects all globes, when it should be limited to the globe targeted by the mouse/touch input.

To reproduce this issue, run WebWorldWind's [MultiWindow example](https://files.worldwind.arc.nasa.gov/artifactory/apps/web/examples/MultiWindow.html) and attempt to navigate on a single globe using either a mouse or a touch screen. The globe that's clicked on or touched is expected to move, while the other two globes remain unchanged.

pdavidc:
- Filtered GestureRecognizer events based on the target WorldWindow
- Fixes #495

### DIFF
--- a/src/BasicWorldWindowController.js
+++ b/src/BasicWorldWindowController.js
@@ -120,8 +120,11 @@ define([
                     this.handleWheelEvent(e);
                 }
                 else {
-                    for (var i = 0; i < GestureRecognizer.allRecognizers.length; i++) {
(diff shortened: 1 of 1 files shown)


pdavidc:
@markpet49 What are your thoughts on this fix for multi-window navigation?

Issue #489, opened 2018-02-23
Screen credit enhancements


pdavidc:

Beak-man:
[Screen credits customization conclusions.pdf](https://github.com/NASAWorldWind/WebWorldWind/files/1948518/Screen.credits.customization.conclusions.pdf)

Issue #494, opened 2018-02-28
Incorrect DrawContext.currentLayer usage


markpet49:
### Description

A number of classes incorrectly rely on dc.currentLayer during the ordered renderable drawing phase (Polygon, AbstractMesh, etc). They should instead rely on the layer property that is captured during the layer drawing phase. See the captured slack notes below:

### Additional Information

@markpet49:                
It would appear to me that AbstractMesh needs to account for the fact that it might not belong to a layer in the opacity calculations at around line 357 in AbstractMesh.doRenderOrdered. Am I correct in assuming that when a mesh is drawn as an ordered renderable that currentLayer is not valid? I believe this works by accident currently because WorldWindow.drawLayers leaves a dangling reference to currentLayer, so the mesh winds up using the opacity from the last layer drawn (usually ViewControlsLayer). In Miguels version, I clear currentLayer at the end of drawLayers which exposed the problem. Is my logic correct on this or is there some subtlety that I'm missing? Thanks.

@pdavidc:
Yes, you're correct, and thanks for cleaning up currentLayer at the end of drawLayers.
• DrawContext.currentLayer is defined during layer rendering, e.g. Layer.render and Renderable.render.
• DrawContext.currentLayer is undefined during ordered rendering. • Shapes keep track of the last layer they encountered during layer rendering.

pdavidc:
@markpet49 An alternate method for fixing this is to ensure that dc.currentLayer is defined during ordered rendering. We could capture the current layer in addOrderedRenderable, then set it appropriately for each entry during ordered rendering. We do this now for globe state.

markpet49:
When fixing this issue, please also address the TODO at the bottom of WorldWindow.drawLayers:
            // TODO: Following line commented out as workaround until issue #494 is addressed.
            // dc.currentLayer = null;

pdavidc:
@zglueck This may be related to the recent work done to support layer opacity in shapes. Would you please look at usages of DrawContext.currentLayer and identify the places that are currently problematic?

zglueck:
@markpet49  and @pdavidc I've reviewed uses of DrawContext.currentLayer and found some usage during ordered rendering operations, which exposes potential issues if the dangling layer reference does not correspond to the actual renderables layer.

All instances of invalid currentLayer use share a common pattern of actually assigning the current layer (during the render call) to a layer property attached to the ordered renderable. Some psuedo code to better illustrate:
```
render = function (dc) {
[...]

PR #499, opened 2018-03-02
Refactored ScreenCreditsController to use existing components

### LINKED ISSUE #467: Refactor ScreenCreditController to reuse existing components
Closes #489

Refactor the ScreenCreditController's drawing code to reuse existing drawing components, such as ScreenImage and ScreenText.

Beak-man:
### Description of the Change
- The current `ScreenCreditsController` makes use of its own facilities to render images and text overlays in the viewport. This change enables it to use `ScreenText` and `ScreenImage` instead.
- Makes the `ScreenCreditsController` to operate as an internal layer.

- It reuses present resources (the two aforementioned classes) and it sets the stage for future enhancements in regards to screen credits as described in the issues documented in epic #489.

### Benefits
 - It greatly simplifies the controller, making it easier to maintain. 
 - Code is reused instead of having redundant rendering facilities. 
 - Sets the stage for enhancing screen credits capabilities, as described in issues related to epic #489.
 - The concept of an internal layer may be useful for other features in the future.

### Potential Drawbacks
- The changes done to enable `ScreenCreditsController` to function as an internal layer, exposed a bug described in #494. `WorldWindow`'s line 1203 was commented out as a temporary workaround.
- Maybe moving `ScreenCreditsController` to the /Layer folder should be considered.
- As for now, a fixed spacing between text and image credits is used. It operates well with current credits as requested by `OpenStreetMapLayer` and `BingTiledImageLayer`, but taller image credits will overlap. This is intended to be addressed in a future PR.

### Applicable Issues
Closes #467


pdavidc:
@Beak-man A brief scan showed changes to files unrelated to the ScreenCreditController refactoring. Please remove those changes. I'd like to see pull requests that are specific to the subject at hand.

That said, I see you're taking the initiative to correct equality and inequality operators that are at risk of type coercion. This is great! Make those changes in another branch, and submit a separate pull request. You can do this in the future as well.

Beak-man:
Will do @pdavidc !

pdavidc:
@Beak-man this refactoring of ScreenCreditController is wonderfully simple and straightforward. It exemplifies the type of consolidation and simplification we're looking for. 

I have a few comments that I think will polish what you've done, and ought to be simple to implement. I've placed those comments inline.

src/render/ScreenCreditController.js
```diff
@@ -51,41 +52,36 @@ define([
          * @classdesc Collects and displays screen credits.
          */
         var ScreenCreditController = function () {
+            RenderableLayer.call(this, "ScreenCreditController");
```
pdavidc: RenderableLayer is a very sensible place to start for a Layer subclass. That said, I don't see how we're benefiting from it. We already manage all the renderable, and override doRender.

src/render/ScreenCreditController.js
```diff
@@ -51,41 +52,36 @@ define([
...
             // Internal. Intentionally not documented.
-            this.imageUrls = [];
+            this.imageCredits = [];
 
             // Internal. Intentionally not documented.
-            this.stringCredits = [];
+            this.textCredits = [];
 
             // Internal. Intentionally not documented.
             this.imageCreditSize = 64;
```
pdavidc: This property is no longer used.

src/render/ScreenCreditController.js
```diff
@@ -121,184 +121,34 @@ define([
...
         // Internal use only. Intentionally not documented.
-        ScreenCreditController.prototype.endDrawingCredits = function (dc) {
-            var gl = dc.currentGlContext,
-                program = dc.currentProgram;
-
-            // Clear the vertex attribute state.
-            gl.disableVertexAttribArray(program.vertexPointLocation);
-            gl.disableVertexAttribArray(program.vertexTexCoordLocation);
+        ScreenCreditController.prototype.doRender = function (dc) {
+            var j = 1;
```
pdavidc: This property is in need of a more descriptive name. Perhaps creditCount or creditOrdinal.

src/render/ScreenCreditController.js
```diff
@@ -121,184 +121,34 @@ define([
...
-            // Scale the image to fit within a constrained size.
-            imageWidth = activeTexture.imageWidth;
-            imageHeight = activeTexture.imageHeight;
-            if (imageWidth <= this.imageCreditSize && this.imageHeight <= this.imageCreditSize) {
-                scale = 1;
-            } else if (imageWidth >= imageHeight) {
-                scale = this.imageCreditSize / imageWidth;
-            } else {
-                scale = this.imageCreditSize / imageHeight;
+            for (var i = 0; i < this.imageCredits.length; i++) {
```
pdavidc: Unintentional O(n) loop. Array.length can be, last I checked, a linear-time operation. The pattern we use throughout WorldWind, though it's not obvious at first glance, is as follows:
`for (var i = 0, len = this.imageCredits.length; i < len; i++) {`

test/BasicWorldWindowController.test.js
```diff
@@ -54,6 +54,7 @@ define([
             return {left: 339.5, top: 225};
         }
     };
+    wwd.layers = [];
```
pdavidc: Would you please clarify why this change is needed in the tests?

src/render/ScreenCreditController.js
Beak-man: Interesting. Will do.

test/BasicWorldWindowController.test.js
markpet49: @pdavidc I added this to fix a unit test breakage when we converted ScreenCreditController to a layer.

markpet49: Also, I moved the Mock WWD creation code to a common routine in the camera branch, so these three changes will all be consolidated.

pdavidc: Thanks for clarifying, Mark.

src/render/ScreenCreditController.js
```diff
@@ -18,74 +18,58 @@
  */
```
markpet49: @pdavidc Should we leave ScreenCreditController in the render folder given that it is now a layer? Given that it is the first of WWW's internal layers, perhaps putting it in the layer folder sends the wrong message.

pdavidc: I agree with your thinking @markpet49. At the moment, the layer folder is intended for application focused layers.

src/render/ScreenCreditController.js
```diff
@@ -121,184 +108,33 @@ define([
...
-            program.loadOpacity(gl, this.opacity);
-
-            ScreenCreditController.texCoordMatrix.setToIdentity();
-            ScreenCreditController.texCoordMatrix.multiplyByTextureTransform(activeTexture);
-            program.loadTextureMatrix(gl, ScreenCreditController.texCoordMatrix);
-
-            if (activeTexture.bind(dc)) { // returns false if active texture cannot be bound
-                // Draw the image quad.
-                gl.drawArrays(gl.TRIANGLE_STRIP, 0, 4);
+            for (i = 0; i < this.textCredits.length; i++) {
```
pdavidc: Two problems here:
• Array.length is called every iteration.
• Variable i should be declared at the top of the method. Avoid dependencies on variable hosting.

Beak-man: Whoops on that Array.length call.
Both points are corrected now.

markpet49:
@Beak-man It looks like there's an off by one type of error with respect to the Y coordinate placement of the credits. The attached image shows the dev version overlayed with your version semi-transparently. If look at the pixel placement, it looks like the new version is slowly rising upward by increasing amounts on the Y axis. Can you look into this?

markpet49:
credits

Beak-man:
@markpet49 , I'll review the accuracy. Is it worth it, though? The placement in this proposal is hardcoded, as we discussed now, and it doesn't take into account the actual heights of the textures.

Beak-man:
@pdavidc , I'll check that out. It never occurred to me, and I didn't identify how the controller was avoiding this repetition before.

markpet49:
@Beak-man Ah, so the difference is the result of a change in the placement algorithm. If so, then never mind.

Beak-man:
@markpet49 , yes. Previous attempts replicated 1:1 credit placement with how the controller worked before, but that relied on reading the credit ActiveTexture, which then we arranged to avoid. It seems that I'm just adding a pixel with each credit, though. That seems a quick fix.

src/render/ScreenCreditController.js
```diff
@@ -121,184 +115,42 @@ define([
...
-
-            for (var i = 0; i < this.imageUrls.length; i++) {
-                imageHeight = this.drawImageCredit(dc, this.imageUrls[i], imageX, this.margin);
-                if (imageHeight > 0) {
-                    imageX -= (this.margin + this.imageCreditSize);
-                    maxImageHeight = WWMath.max(imageHeight, maxImageHeight);
+            // Verify if text credit is not already in controller, if it is, don't add it.
+            for (var i = 0, len = this.textCredits.length; i < len; i++){
+                if (this.textCredits[i].text === stringCredit){
+                    return 0;
```
pdavidc: Just `return;` here.

Beak-man: Rght. Bad habits die hard.

markpet49:
Looks good to go....

PR #502, opened 2018-03-06
Allow setting a canvas element instead of the ID of a canvas when creating a WorldWindow

### LINKED ISSUE #479: Support HTMLCanvasElement object as argument of the WorldWindow constructor
### Description

The proposition is to support giving an HTMLCanvasElement object to the constructor of WorldWindow as alternative to giving the ID of an existing canvas. This would greatly facilitate the integration of WebWorldWind with modern JavaScript frameworks like React, Angular and Polymer, which manages the DOM elements themselves.

**Expected behavior:**
a) Giving a string to WorldWindow uses getElementById for locating the actual canvas.
b) Giving an HTMLCanvasElement object to WorldWindow uses this canvas directly for the rendering.
c) Giving anything else to WorldWind fails.

AkeluX:
Closes #479

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -2136,6 +2136,16 @@
             "tweetnacl": "0.14.5"
           }
         },
+        "string-width": {
+          "version": "1.0.2",
+          "bundled": true,
+          "dev": true,
+          "requires": {
+            "code-point-at": "1.1.0",
+            "is-fullwidth-code-point": "1.0.0",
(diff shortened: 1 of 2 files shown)


pdavidc:
@markpet49 Do you have any thoughts or questions about this change?

AkeluX:
@markpet49 Absolutely. It's done.

PR #532, opened 2018-03-13
Improve GPU resource cache management

### LINKED ISSUE #497: Blank tiles appear after viewing large amounts of imagery
### Description

Blank tiles and poor performance experienced after viewing large amounts of imagery.  This problem is manifest in full screen desktop displays (landscape).

### Steps to Reproduce

This issue can be replicated in the BasicExample using the OSM layer by creating a large viewport.  

Change the BasicExample to use a full width and height canvas: change the `div.container` class to `container-fluid`, and change the `div#globe.col-sm-9` class to `col-sm-12`. Also change the canvas height from `auto` to `calc(100vh)`.
Example:
```
<div class="container-fluid">
    <div class="row">
        <div class="col-sm-12">
            <h4>Layers</h4>
            <div class="list-group" id="layerList">
            </div>
        </div>
        <div class="col-sm-12" id="globe">
            <canvas id="canvasOne" width="1000" height="1000" style="width: 100%; height: calc(100vh)">
                Your browser does not support HTML5 Canvas.
            </canvas>
        </div>
    </div>
</div>
```
1. Zoom in close to terrain
2. Tilt the view so horizon is in view
3. Using view controls, rotate the view continuously, expecting to see blank tiles within previously viewed content
4. Zoom out, expecting to see blank tiles in previously viewed content
5. Repeat 1 through 4. Over time the UI performance lags, possibly resulting in a lockup on the web page; and zoomed-out views fail to repopulate tiles.

- No loss of performance
- Scene quickly and completely repopulated

- Blank tiles appear that are either slow or fail to repopulate
- Globe navigation becomes sluggish
- Continuous network activity as tiles are re-downloaded 

**Reproduces how often:** [What percentage of the time does it reproduce?]
100%

Windows 10

### Additional Information

Any additional information, configuration or data that might be necessary to reproduce the issue.

emxsys:
### Description of the Change
Adds an interface to GpuResourceCache and its backing MemoryCache to advance or retard the normal aging of a cache entry.  Typically users of the cache would an _aging multiplier_ on volatile cache entries so they age faster and evict sooner.

Provides desirable and expected behaviors when viewing the globe such that recently viewed map tiles remain in the cache in preference to temporal text entries like view coordinates.

### Benefits
- Provides an orderly eviction of volatile resources.
- The resource eviction policies are under the control the resource creators. 
- The existing GpuResourceCache and MemoryCache function interfaces are unchanged.

### Potential Drawbacks
- A 100x aging multiplier has been applied to text textures.
- A 10x aging multiplier has been applied to surface shape tiles.

### Applicable Issues
Closes #497

### DIFF
--- a/src/WorldWindow.js
+++ b/src/WorldWindow.js
@@ -73,25 +73,33 @@ define([
          * @alias WorldWindow
          * @constructor
          * @classdesc Represents a WorldWind window for an HTML canvas.
-         * @param {String} canvasName The name assigned to the HTML canvas in the document.
+         * @param {String|HTMLCanvasElement} canvasElem The ID assigned to the HTML canvas in the document or the canvas
+         * element itself.
          * @param {ElevationModel} elevationModel An optional argument indicating the elevation model to use for the World
          * Window. If missing or null, a default elevation model is used.
-         * @throws {ArgumentError} If there is no HTML element with the specified name in the document, or if the
+         * @throws {ArgumentError} If there is no HTML element with the specified ID in the document, or if the
          * HTML canvas does not support WebGL.
          */
-        var WorldWindow = function (canvasName, elevationModel) {
+        var WorldWindow = function (canvasElem, elevationModel) {
             if (!(window.WebGLRenderingContext)) {
                 throw new ArgumentError(
                     Logger.logMessage(Logger.LEVEL_SEVERE, "WorldWindow", "constructor",
                         "The specified canvas does not support WebGL."));
             }
 
-            // Attempt to get the HTML canvas with the specified name.
-            var canvas = document.getElementById(canvasName);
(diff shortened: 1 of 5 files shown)


emxsys:
Debugging code is available in the MemoryCache for the testing. Set DEBUG=true to dump the MemoryCache to the console during evictions.

pdavidc:
@emxsys and I just discussed this: There's a minor naming change to make, but otherwise this looks good to me.

emxsys:
Updated pull request to use 'agingFactor' term and removed debug code.  Tested with Shapefiles example to assert correct behavior with Text, ShapeSurfaceTiles, and TiledImageLayer.

pdavidc:
Looks good to me now.

PR #546, opened 2018-03-15
Fix layer reference for ordered rendering

### LINKED ISSUE #494: Incorrect DrawContext.currentLayer usage

zglueck:
### Description of the Change

This changes the ordered renderables use of DrawContext's current layer. DrawContext.currentLayer may be invalid during ordered rendering, the reference has been replaced to an object property of layer assigned during rendering.

Additionally, this PR removes the layer reference following layer rendering.

### Applicable Issues

Closes #494

### DIFF
--- a/src/WorldWindow.js
+++ b/src/WorldWindow.js
@@ -1207,8 +1207,7 @@ define([
                     }
                 }
             }
-            // TODO: Following line commented out as workaround until issue #494 is addressed.
-            // dc.currentLayer = null;
+            dc.currentLayer = null;
             var now = Date.now();
             dc.frameStatistics.layerRenderingTime = now - beginTime;
         };
--- a/src/formats/collada/ColladaScene.js
+++ b/src/formats/collada/ColladaScene.js
@@ -668,7 +668,7 @@ define([
             }
 
             var color = new Color(r, g, b, a);
-            opacity = a * dc.currentLayer.opacity;
+            opacity = a * this.layer.opacity;
             gl.depthMask(opacity >= 1 || dc.pickingMode);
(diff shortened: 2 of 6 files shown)


pdavidc:
@zglueck The changes look good. Have each of the affected shapes been tested for correctness in picking, opacity, and anything else that might relate to currentLayer?

zglueck:
@pdavidc testing showed everything working as expected.

PR #548, opened 2018-03-15
Modified BoundingBox.setToSector to handle large sectors

### LINKED ISSUE #545: Top level tile for a WMTS CRS3857 layer has no volume
### Description
A WMTS CRS 3857 layer of US national forests was disappearing and reappearing as I panned and tilted the globe.  I traced the code and discovered that sometimes the view frustum didn't intersect the top level tile(s). When this occurred the layer wouldn't be displayed.  I rendered the tile extents (BoundingBox) and discovered the top level tile has no volume.

The layer that manifests this issue has a single top level tile that spans the hemispheres (+/-180 longitude, +/-85 latitude).

### Illustration
Here are three coincident views of the globe to document the bounding box problem. The first two images are for orientation, the third image documents the problem.

The following image is for reference and orientation. It shows the BlueMarble layer and a WMTS CRS 3857 overlay of US national forests.
image

 ---

The following image shows just the WMTS layer of US national forests 
image

---

This image shows the bounding boxes (volumes) for the WMTS layer. Note the level 0 bounding box appears to have no volume.
image

pdavidc:
- Sectors with a longitude span greater than 180 degrees are correctly represented
- Closes #545

### DIFF
--- a/src/geom/BoundingBox.js
+++ b/src/geom/BoundingBox.js
@@ -318,6 +318,16 @@ define([
                 this.adjustExtremes(this.r, rExtremes, this.s, sExtremes, this.t, tExtremes, this.tmp1);
             }
 
+            // If the sector encompasses more than one hemisphere, the 3x3 grid does not capture enough detail to bound
+            // the sector. The antipodal points along the parallel through the sector's centroid represent its extremes
+            // in longitude. Incorporate those antipodal points into the extremes along each axis.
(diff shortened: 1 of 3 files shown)


pdavidc:
FYI, This pull request depends on the fix in #547, which I chose to separate in order to highlight the issue. The Travis CI build will pass once #547 is merged.

PR #550, opened 2018-03-19
Enhancement/elevation coverages

### LINKED ISSUE #422: Compound elevation models

markpet49:
### Description

Notes regarding how this PR addresses points in the issue are inline below.

> 1. getTimestamp actually returns the max timestamp, i.e. allows to get the timestamp of the newest data in the model. It might be useful to get the timestamp of the oldest data in the model as well. I would therefore have getMaxTimestamp and getMinTimestamp.

ElevationModel and Globe now split the timeStamp property into min/max versions. 

> 2. The definition of hasCoverage(Sector) should be clarified between "has data for at least one point" and "has data for all the point of the sector". It seems that the former is what is implement, e.g. in TiledElevationCoverage. Would two methods be useful?

Not addressed as directed below. 

> 3. Something I don't understand, but also haven't had time to look deeper at, is what is the resolution of the result array and whether this has to be coordinated somehow between the various coverages.

Not addressed.

> 4. The priority between the coverages seems to be the order of the ArrayList. The last item can override whatever it wants in the result array. I think this is fair enough, but we might want to offer the possibly to insert coverages at any place in the ArrayList, not only at the end with addCoverage and addAllCoverages. This would be the equivalent of insert in ArrayList.

ElevationModel now implements a sortCoverages property. By default, this property is true and ElevationModel maintains the coverages in sorted order, such that the highest resolution data is preferred, if available. 

If sortCoverages is false, the coverage order is maintained in the order in which the application adds the coverages and can be re-ordered using insertCoverage and removeCoverageAtIndex.

> ElevationCoverage's displayName and enabled properties, like their Layer counterparts, provide essential functionality for applications.

These properties are implemented in ElevationCoverage and respected by ElevationModel.

> ElevationCoverage's user property interface are non-essential, as are the corresponding convenience methods on ElevationModel. These are part of a broader app convenience interface on layers and elevations, which we don't have in WebWorldWind yet.

The user property interface was not replicated. 

> The timestamp methods are intended as a stopgap measure to control terrain regeneration. It's not a key part of the interface concept, though we need it as we have no alternatives at the moment.

No action. 

> The hasCoverage methods turned out to be non-essential, and I removed them in an iteration after receiving @AkeluX's review.

hasCoverage is not implemented.

> ElevationModel may make more sense as a self ordering set, rather than an ordered list. Applications often aren't able to discern the order in which elevations ought to appear. ElevationModel's methods ought to reflect whatever choice we make here.

The current implementation orders by resolution by default, but the sorting can be turned off. 

> The getHeight methods may be more clear if they provide control over the desired resolution and indicate the resolution of the result array.

ElevationModel.elevationsForGrid returns resolution information, ElevationModel.elevationsAtLocation does not. There would be some additional refactoring required to make elevationsAtLocation which may make more sense as a separate issue. 

### Applicable Issues

Closes #422


markpet49:
@zglueck Hi Zach, I included you as a reviewer on this as I had to move a change or two that you made when I merged with develop. Let me know if I broke any of your stuff.

pdavidc:
@markpet49 Nice work! I've added additional comments in #422 (rather than here) to keep our general notes on compassable elevations in one place.

markpet49:
Closed pending further updates.

PR #576, opened 2018-03-21
Addition of screen credit hyperlink support

### LINKED ISSUE #475: Screen credit hyperlink support

Beak-man:
### Description of the Change
Enables screen credits to work as hyperlinks via the following changes:
- Click handling in `BasicWorldWindowController`.
- Addition of a new optional hyperlink parameter to `ScreenCreditController.addStringCredit` and `ScreenCreditController.addImageCredit`.
- Appending a `url` object to the `userProperties` of each credit that's intended as a hyperlink.

- Enables use cases where screen credits aid in attributions by optionally operating as hyperlinks.
- Implements general click handling in `BasicWorldWindowController` that may  be expanded upon for other use cases.

### Potential Drawbacks
- It enables picking to `OpenStreetMapImageLayer`, `DigitalGlobeImageLayer`, and `BingTiledImageLayer` as a workaround required to enable picking to the screen credits. `ScreenCreditController` itself works as an internal layer that has `pickEnabled` set to true, but due to the fact that the credits don't persist and rely on the host layer to add them on every render, the host layers themselves require to have picking enabled.
- Click handling in `WorldWindowBasicController` is very specifically tied to reading an assumed `pickedObject.userObject.userProperties.url` object. A more generic solution for click handling inside this controller can be devised at a later stage.
- Mouse cursor doesn't change when hovering over hyperlink, as is the case with the usual browser DOM. Therefore there's no UI metaphor to rely to the user which screen credits operate as hyperlinks and which ones are not before clicking on them.

### Applicable Issues
Closes #475


src/layer/BingTiledImageLayer.js
```diff
@@ -47,19 +49,22 @@ define([
...
+            this.creditImage = WorldWind.configuration.baseUrl + "images/powered-by-bing.png";
         };
 
         BingTiledImageLayer.prototype = Object.create(MercatorTiledImageLayer.prototype);
 
         BingTiledImageLayer.prototype.doRender = function (dc) {
             MercatorTiledImageLayer.prototype.doRender.call(this, dc);
             if (this.inCurrentFrame) {
-                dc.screenCreditController.addImageCredit(this.creditImage);
+                dc.screenCreditController.addStringCredit("Imagery courtesy of Bing maps", Color.DARK_GRAY, "https://www.bing.com/maps");
```
Beak-man: Hyperlinks added to Bing layer for the purposes of having a functional test for this PR.

src/layer/DigitalGlobeTiledImageLayer.js
```diff
@@ -95,8 +96,12 @@ define([
             this.urlBuilder = {
                 urlForTile: function (tile, imageFormat) {
                     if (!self.metadataRetrievalInProcess) {
-                        return self.urlTemplate.replace("{z}", (tile.level.levelNumber + 1)).
-                            replace("{x}", tile.column).replace("{y}", tile.row);
+                        return self.urlTemplate.replace(
```
Beak-man: Webstorm's code reformatting tool was putting all these arguments in a single line that went beyond the recommended width so I manually arranged them in this arguably more readable manner.

pdavidc:
@Beak-man @markpet49 @zglueck 

@Beak-man and I just reviewed this change over a Slack call. Here's a summary of the conversation:
- Enabling picking on TiledImageLayer does not adversely affect picking behavior. TiledImageLayer does draw image tiles during picking, but does so without affecting the pick frame. This is a performance concern that we'll address separately, but not a functional concern.
- BasicWorldWindowController's tap/click test ought to be testing the top picked object, not the entire list.
- ScreenCreditController needs some minor formatting adjustments.

pdavidc:
@Beak-man What's wrong with the form I showed yesterday?

```javascript
var topObject = pickList.topPickedObject();
if (topObject.userObject && 
    topObject.userObject.userProperties && 
    topObject.userObject.userProperties.url) {
    // the top picked object has a url
}
```

Beak-man:
Now I see what I was doing wrong. I was checking for the object chain after `topObject`, but wasn't assuring the the existence of `topObject` itself first. Pushing the correction now.

markpet49:
@zglueck Any concerns with this PR?

zglueck:
@markpet49  and @Beak-man Just a question, why are the userProperties being used to store the url instead of an instance property?

pdavidc:
While we could append a property to the Renderable objects representing screen credits, their existence is undocumented. userProperties is designed to handle loose coupling between a data model and a controller, in cases like this.

For our team's consideration when we look at next steps, what ideas come to mind @zglueck?

zglueck:
I'm not as familiar as I would like to be with this framework to provide a substantive next step. My reason for asking was concern with the library adding properties to an object described as "an application defined object". I'm just afraid we might be stepping on toes.

So, I come to the party without an hors d'oeuvres, but complain about the drapes :)

PR #578, opened 2018-03-26
Accept plain JS objects as data source for the GeoJSON parser

strikerM:
### Description of the Change
Allows the GeoJSONParser to receive a plain js object as the data source, in addition to a url or a JSON string

The change is in a core class

### Benefits
Applications or other WorldWind components, such as the OpenSearchLayer, don't need to serialize js objects before passing them to the GeoJSON parser

### Potential Drawbacks
Objects are passed by reference, changes to a feature's properties will be reflected in the renderable userProperties


AkeluX:
@strikerM I've simplified the execution flow for the possible data source types. What do you think of these changes?

strikerM:
Looks good to me.

PR #580, opened 2018-03-26
Enhancement/elevation coverages

markpet49:
### Description of the Change
The first iteration of changes required to move toward fulfilling the requirements of issue #422. Responses to various issue comments below:

> getTimestamp actually returns the max timestamp, i.e. allows to get the timestamp of the newest data in the model. It might be useful to get the timestamp of the oldest data in the model as well. I would therefore have getMaxTimestamp and getMinTimestamp.

The decision was to go with a single timestamp representing the maximum elevation time.

> The definition of hasCoverage(Sector) should be clarified between "has data for at least one point" and "has data for all the point of the sector". It seems that the former is what is implement, e.g. in TiledElevationCoverage. Would two methods be useful?

hasCoverage was determined to be unnecessary.

> Something I don't understand, but also haven't had time to look deeper at, is what is the resolution of the result array and whether this has to be coordinated somehow between the various coverages.

In the new implementation, data from the highest resolution coverage is always preferred. Should data from the highest coverage be unavailable, data from the next highest resolution coverage is used and so on until the data request is filled or we run out of coverages.

It is possible that a result set of elevation data will contain data of various resolutions. This will most often occur at the edges of a coverage's bounding sector, but can occur anywhere.

In a future iteration, it will be possible to request a preferred resolution and have data from that resolution preferred over higher resolution data.

> The priority between the coverages seems to be the order of the ArrayList. The last item can override whatever it wants in the result array. I think this is fair enough, but we might want to offer the possibly to insert coverages at any place in the ArrayList, not only at the end with addCoverage and addAllCoverages. This would be the equivalent of insert in ArrayList.

In the new implementation, data from the highest resolution coverage is always preferred. In a future iteration, it will be possible to request a preferred resolution and have data from that resolution preferred over higher resolution data.

> ElevationCoverage's displayName and enabled properties, like their Layer counterparts, provide essential functionality for applications.

These properties are fully supported.

> ElevationCoverage's user property interface are non-essential, as are the corresponding convenience methods on ElevationModel. These are part of a broader app convenience interface on layers and elevations, which we don't have in WebWorldWind yet.

User properties are not supported.

> The timestamp methods are intended as a stopgap measure to control terrain regeneration. It's not a key part of the interface concept, though we need it as we have no alternatives at the moment.

Timestamp support has been retained.

> The hasCoverage methods turned out to be non-essential, and I removed them in an iteration after receiving @AkeluX's review.

hasCoverage is not implemented.

> ElevationModel may make more sense as a self ordering set, rather than an ordered list. Applications often aren't able to discern the order in which elevations ought to appear. ElevationModel's methods ought to reflect whatever choice we make here.

> The getHeight methods may be more clear if they provide control over the desired resolution and indicate the resolution of the result array.

> Java's initial compound elevation system was an ordered list, like Android's is today. This was simple, but left it up to the app to order coverages, and they often didn't realize they needed to manage this. I've found that most developers don't think about coverages as composable, so the topic of composition and ordering confuses them.

>Our next iteration for Java added self-ordering based on the coverage resolution. This worked well, but its implementation had a subtle bug which frustrated users, and gave app developers no recourse: coverage resolution was considered global for the purposes of ordering. This simplification broke down with our default coverage, which is multi-resolution: 900m oceans, 30m continents, 10m USA. For example, an app could add 100m bathymetric data, but the sorting step treated the lower-resolution default coverage as 10m, resulting in a sort order that never displays the custom app coverage.

The decision was to go with simple resolution ordering in this iteration. Support for application controlled resolution preferences will be provided in a future iteration.

> Ordering cannot be global (see my previous comment on Mar 19).

In the current implementation, higher resolution is preferred, but lower resolution data will be used to fill the request if the higher resolution data is unavailable.

> Querying an elevation at location can be done in reverse order to eliminate redundant lookups.

Fixed.

> Querying an elevation grid requires a step to handle locations with NODATA values and locations outside the coverage area. In Android, this is implemented by letting the coverages write directly to the result array.

Addressed.

> Testing elevations against 0 is not a reliable indicator of existence of value. Identifying a standard NODATA value is required, or a separate test (though the latter is much less efficient). Options are NaN, null, undefined.

Fixed, NaN is used.

> Priority based on sector ought to be sufficient. You and I may want to work through the cases involved.

The decision was to go with simple resolution ordering in this iteration. Support for application controlled resolution preferences will be provided in a future iteration.

> I looked into this briefly. I think we should consider taking it up after the merge, as it requires significant changes to TiledElevationCoverage.
I think what works best is to define a standard value we adopt throughout WorldWind. Components reading elevation images can be responsible for converting to that standard when needed. We went through several very problematic versions of this in WWJava. In the end, we realized that nodata values ought to be easy to handle, both in our components and by the app, and there ought to be a single representation throughout.

This comment was regarding NODATA handling. In this iteration, the NODATA value is NaN and each coverage can write to the result array where NODATA elements are present. The net effect is that the elevation results can be multi-resolution with preference given to higher resolutions.

> New topic ZeroElevationModel is obsolete now, correct? I think it's equivalent to an empty ElevationCoverage.

ZeroElevationModel has been removed and the ElevationCoverage base class behaves as a zero elevation coverage.

> New topic WorldWindow could accept an ElevationCoverage, simplifying app transition to 0.10.0 and simplifying the constructor. I think of ElevationModel like LayerList, in that it organizes app components, and an app is unlikely to replace it.

This requirement is not applicable given later design changes.

> Require ElevationCoverage to be single resolution, and drop the notion of a multi-resolution EarthElevationCoverage.

Done.

> Recast EarthElevationCoverage as EarthElevationModel, which will be a shorthand constructor for an ElevationModel containing three independent WMS-based coverages: GEBCO, Aster, USGS NED.

Done.

> Recast EarthRestElevationCoverage as EarthRestElevationModel.

It was decided to leave this as a coverage.

> Recast WcsEarthElevationCoverage as WcsEarthElevationModel and deprecated it. It isn't useful, and WMS vs. WCS coverages will become an implementation detail of our default coverages.

This has been left as a deprecated coverage.

> Order coverages exactly as in the initial implementation, but replace ElevationCoverage's getBestResolution with a property indicating the sole resolution.

Done.

> In elevationsForGrid, optimize for multiple coverages and return an array containing the coverages that contributed to the result. We'll need to discuss what optimizing for multiple coverages might look like separately.

Done.

> Our current method of a serving elevations in a single, multi-resolution elevation coverage is inherently problematic, and complicates WorldWind's elevation interfaces. It's important that we support retrieval of elevations from WorldWind's existing WMS-based elevation services, but we have the flexibility to change how we retrieve those elevations. Said another way, we currently make WMS requests for the layers "GEBCO,aster_v2,USGS-NED", but we can also request those same layers independently. It means more HTTP traffic, but greatly simplifies WorldWind's elevation interfaces, and provides a correct and reliable solution for composing application elevations.

> With regard to correctness, consider an application coverage with a resolution between Aster and USGS NED. The mechanism I mentioned on Mar 19: ordering coverages based on sector, still fails when the sector contains Aster, USGS NED, and the app coverage: the app coverage should be ordered between Aster and USGS NED, but there's no way to achieve that. So the app coverage would appear correctly ordered, but only when the terrain tile in question intersected only the Aster sector. The concept of a single, multi-resolution coverage amounts to a nested ElevationModel. Since we don't plan to implement a hierarchical elevation model, I think a clear choice is to separate the model's coverages.

The coverages have been split out.

> Modify elevationsForGrid to identify the coverage with resolution closest to the specified target, then attempt to satisfy the request with the target coverage.
If the target coverage cannot satisfy the entire request, iterate over lower-resolution coverages (if available) to fill in the blanks.

Targeting a coverage resolution will be addressed in the next iteration.

> Coverages are ordered with respect to resolution. Given the potential benefit, it's unclear to me whether application control over coverages ordering is desirable.
Coverages satisfy the result where they can, and indicate whether the result is complete or partial.
ElevationModel performs the composition of multiple coverage responses.

Support for unsorted coverages has been removed. Composition from multiple coverages has been implemented.

> Provided we implement the changes in my previous comment on Mar 20, multiple coverages will be the common case for WorldWind. Adjusting ElevationModel's implementation as I've outlined should maintain adequate performance, and has the side effect of increasing compatibility with non-WorldWind elevation services.

Performance testing indicates good performance.

> Nearly every elevation service we encounter is not designed with a client like WorldWind in mind. These services lack any sort of caching capability, and perform very until poorly when client requests resolutions are vastly different from the underlying data model. Said another way, these services tend to fail or respond slowly to WorldWind's top level tile requests (geographically-large and low-resolution request), and lack the caching that would eventually stabilize the service. The change to request only what's nearest to the desired resolution means WorldWind is more likely to make requests that the service can respond to in a timely manner.

This comment will be addressed in the next iteration when targeted resolutions are supported.

### Applicable Issues

#422


pdavidc:
@markpet49 Very thorough work!

Regarding ZeroElevationModel, my design goal is that ZeroElevationModel is replaced by an *empty* ElevationModel. So initializing a single no-op coverage to set up Globe2D would be unnecessary. From what I've seen, ElevationModel is ready to do this.

Regarding primitive arguments, we skip any null/undefined argument checking on these. So elevationAtLocation need not check anything. This was a conscious performance optimization made in the project's early days.

markpet49:
@pdavidc Thanks Dave, I will adjust the code and let you know when it's ready for another look.

markpet49:
@pdavidc The above adjustments have been made. Globe2D works just fine with an empty elevation model.

src/globe/ElevationCoverage.js
```diff
@@ -0,0 +1,138 @@
...
+             * @default "Elevations"
+             */
+            this.displayName = "Zero Elevations";
+
+            /**
+             * Indicates whether or not to use this coverage.
+             * @type {Boolean}
+             * @default true
+             */
+            this.enabled = true;
```
pdavidc: Changing a coverage's enabled property does not cause the terrain to regenerate. One possible adjustment is to update the coverage timestamp when the enabled property changes:
      
```javascript
Object.defineProperties(ElevationCoverage.prototype, {
    /**
     * Indicates whether or not to use this coverage.
     * @type {Boolean}
     * @default true
     */
    enabled: {
        get: function () {
            return this._enabled;
        },
        set: function (value) {
            this._enabled = value;
            this.timestamp = Date.now();
        }
    }
});
```

markpet49: @pdavidc Hi Dave. I think all your concerns have been addressed. Notes:

1. The areaElevationsForGrid resultFilled computation was also causing a subtle bug, so I changed it to operate the same way as pointElevationsForGrid as you suggested.
2. I changed all the external facing APIs to use degrees for resolution as we discussed. LevelSets still use radians internally.
3. Coverage resolutions and levels have been updated as we discussed.
4. I removed "Earth" from the coverage names. 
5. I had left removeCoverageAtIndex in place as a convenience method. It has now been removed.
6. I changed elevationsForGrid to accept a plain Array. I had mistakenly thought Float64Array was required based on usage in the code and a bad assumption.
7. I added the property code for ElevationCoverage.enabled as you suggested. At first it didn't work, due to ElevationModel.timestamp respecting the enabled flag, but that was probably a mistake anyway. Terrain now seems to regenerate fine. 
8. I addressed the various terminology and doc tweaks you pointed out.

markpet49:
@pdavidc Hi Dave. I think this is ready for another look. I added the changes related to NED data we discussed yesterday. It seems to work well.

markpet49:
@pdavidc Hi Dave, I think this is ready for another look. As expected, using mins/maxes across all coverages fixed the clipping issues south of Australia.

markpet49:
@pdavidc Hi Dave. I made the min/max loop more permissive. I also noticed that ElevationModel.minMaxElevationsForSector behaved slightly differently than the old version, so I added some additional failover logic to that. Let me know what you think. It tested out fine for me.

markpet49:
@pdavidc With respect to performance, I saw the following (zoom in to Flagstaff, tilt to horizon, and spin 360):
On Dev:

  | Used | Capacity
-- | -- | --
Tessellator.tileCache | 175944 | 1000000
ElevationModel.tileCache | 175944 | 1000000
ElevationModel.imageCache | 175944 | 1000000

With ElevationCoverages:

  | Used | Capacity
-- | -- | --
Tessellator.tileCache | 2610 | 1000000
ElevationModel.coverages.tileCache | 334492 | 4000000
ElevationModel.coverages.imageCache | 10440 | 4000000

markpet49:
@pdavidc I corrected the performance numbers, let me know if I should measure anything else.

Issue #486, opened 2018-02-23
WCS compatible coverage construction


pdavidc:
Child of #482

Modify the common TiledElevationCoverage implementation such that its construction can be easily used with an OGC Web Coverage Service, versions 1.0.0 and 2.0.1.

**Implementation Considerations**

WorldWind's components that interact with OGC services have historically required configuration at construction time, thereby putting the burden of the capabilities handshake somewhere in the application's domain. In my opinion, this has many drawbacks:
• Component configuration is awkward.
• Components are essentially immutable.
• Components are difficult to extend or compose.
• Applications have no reliable mechanism for responding to capabilities handshake failures that result from a poor network connection. 

Given the above, I think a better design is to put the burden of the capabilities handshake on each of WorldWind's OGC components. The component requires only essential information at construction time, and provides mutable properties (and/or methods) for configuration after construction. The capabilities information is retrieved lazily, as-needed, and the component configured appropriately if and when the capabilities response is received. I think the result will be a simple, easy to understand, and easy to extent first-class WorldWind component.

zglueck:
The creation of the WebCoverageService manages version negotiation and parsing of WCS coverages. It provides a generic object which can be utilized by the refactored TiledElevationCoverage to easily and flexibly create elevation coverages based on WCS data providers.

PR #629, opened 2018-04-05
Initial WebCoverageService Interface

zglueck:
### Description of the Change

Adds the initial public API for the WebCoverageService including functions and classes.

### Applicable Issues

#486 and #482


pdavidc:
@zglueck Great iterative first step!

• Does WcsCoverage need to be specific to WCS? Could it be a general Coverage?

• Could WcsCoverage.elevationConfig be a property rather than a method? JavaScript provides getters, so a property gives us the flexibility to return a static object or a dynamic object, where a method does not.

• Why is there more than one argument on WcsCoverageService, and why are there any arguments to the connect method?

• Could WcsCoverageService be a standard method or property, rather than an asynchronous one?

zglueck:
Thanks @pdavidc , here are my thoughts:

1. I suppose it could, but I need some more information on your thoughts on a general use "Coverage".

2. No problem

3. The two arguments in the constructor for WebCoverageService provide enough information for a fully configured service object to function (it knows who to talk to, what coverages are available, version information). The `connect` function provides an ease of use factory method to retrieve and negotiate capabilities and provide, via Promise, a fully configured WebCoverageService. I think the advantage to this approach is the elimination of an asynchronous initialization call with an easy to use and understand API.

4. I'm not following here, will you elaborate?

zglueck:
@pdavidc I think I captured the key points from our discussion. Please review when you get a chance.

zglueck:
@pdavidc this will be a good point to start from, I've got no adjustments to make at this point

pdavidc:
@AkeluX @markpet49 @emxsys This is our first take at a Web Coverage Service interface, and a potential pattern we could adopt for WMS and WMTS. It doesn't show how the coverage is configured, but we can mock some code demonstrating that. Please send any thoughts you have, as we can easily iterate on this interface from this point.

@zglueck Would you please post a Gist showing how we expect this interface to be used?

zglueck:
@AkeluX @markpet49 @emxsys @pdavidc 

A Gist demonstrating the use of the interface:

https://gist.github.com/zglueck/c197f2ccd89370b96727f45035d016b0

pdavidc:
@Akelux thanks for the insightful thoughts. I’m on travel the bulk of this week, so our next opportunity to coordinate is our next scheduled meeting.

The points you raise are certainly valid. We’ll implement against what’s there now, and plan to adjust.

PR #630, opened 2018-04-10
Add WebCoverageService async capabilities negotiation

zglueck:
### Description of the Change

This pr adds WCS document retrieval for initialization of the WebCoverageService. It includes rudimentary version negotiation and DescribeCoverage request metering. To limit the scope of changes, this pr does not include the full initialization of the WebCoverageService, specifically a meaningful creation of the `coverages` array.

This is an early prototype that I'd like to get some feedback on before fully flushing out functionality. To test, I recommend the following three services:
- http://maps.f17d911.rocks/geoserver/wcs (a WWSK 0.6.0 SNAPSHOT version)
- https://worldwind26.arc.nasa.gov/wcs (our elevation server which requires a version negotiation)
- http://sedac.ciesin.columbia.edu/geoserver/wcs (some 200 odd coverages which uses the DescribeCoverage request metering)

The following code should at least demonstrate if the code is working in your environment:
```javascript
var createAndTestService = function (serviceAddress) {
    WorldWind.WebCoverageService.create(serviceAddress)
        .then(function (wcsCaps) {
            console.log(serviceAddress + " has " + wcsCaps.coverages.length + " coverages.");
        })
        .catch(function (e) {
            console.log("This was caught application side catch statement in the " + serviceAddress + "...");
            console.log(e);
        })
};

createAndTestService("https://worldwind26.arc.nasa.gov/wcs");
```

### Applicable Issues

Progress toward #639


src/ogc/wcs/WebCoverageService.js
```diff
@@ -64,7 +104,183 @@ define([
...
+                return false;
+            }
+
+            var version = xmlDom.documentElement.getAttribute("version");
+
+            return WebCoverageService.COMPATIBLE_WCS_VERSIONS.indexOf(version) >= 0;
+        };
+
+        // Internal use only - copied from WmsUrlBuilder, is there a better place to centralize???
+        WebCoverageService.prototype.prepareBaseUrl = function (url) {
```
zglueck: This copied from WmsUrlBuilder. It is statically accessible from WmsUrlBuilder but not documented to be used externally.

I opted for adding it here as an object method to facilitate extension, specifically the future inclusion of an authentication token. I'm concerned about the copy paste code starting to creep throughout OGC requests.

src/ogc/wcs/WebCoverageService.js
```diff
@@ -35,27 +35,67 @@ define([
...
+
+        /**
+         * Contacts the Web Coverage Service specified by the service address. This function handles version negotiation
+         * and capabilities and describe coverage document retrieval. The return is a Promise to a fully initialized
+         * WebCoverageService which includes an array of WcsCoverage objects available from this service.
+         * @param serviceAddress the url of the WebCoverageService
+         * @returns {PromiseLike<WebCoverageService>}
          */
-        WebCoverageService.prototype.connect = function () {
+        WebCoverageService.create = function (serviceAddress) {
```
zglueck: I jumped between `load` and `create` and settled on `create` because of how it looks when a client invokes it. It seems to make sense:
```
WebCoverageService.create("https://worldwind26.arc.nasa.gov/wcs")
    .then(function (wcs) {
       wcs.coverages...
});
```

I think there is a potential for confusion on creating a Web Coverage Service, but discover, init, load, didn't seem to clearly explain we are creating our client side WebCoverageService for interaction with a Web Coverage Service.

src/ogc/wcs/WebCoverageService.js
```diff
@@ -35,27 +35,67 @@ define([
...
+         * An array of compatible WCS versions supported by this client service.
+         * @type {string[]}
+         */
+        WebCoverageService.COMPATIBLE_WCS_VERSIONS = ["1.0.0", "2.0.0", "2.0.1"];
+
+        /**
+         * The maximum length of a constructed DescribeCoverage url request.
+         * @type {number}
+         */
+        WebCoverageService.MAX_URL_LENGTH = 2083;
```
zglueck: This value is somewhat magical. I remembered the WMS issue with too long URLs. I did a quick search and found IE doesn't allow urls longer than 2083 characters. I assumed this was the limitation but after finding the original issue, I found it was actually an Apache limit of 4096. I'm not even sure the 2083 applies to XHR requests in IE. I was ready to change to 4096 but noticed one of my test servers (http://sedac.ciesin.columbia.edu/geoserver/wcs) fails with longer URLs. It works with urls less than 2083 but not with 4096. I don't know where the cutoff is. I retained the maximum however because I believe it is a good practice for larger WCS servers.

AkeluX:
Yes, I believe we can apply the same pattern to the OpenSearch case. We would have the create method retrieve and parse the OpenSearch description document. The created service would then provide search URLs with the relevant configuration like you will provide a specific coverage.

I like `create` better than `connect`. It represents well the fact that we create the service and is generic enough to cover any type of creational activities, e.g. version negotiation, capabilities discovery, etc.

PR #637, opened 2018-04-18
Add an editor for surface shapes

claudiaifrim:
### Description of the Change
Provide the implementation of the Shape Editor. 

It is highly desirable to provide to the users a simple solution to edit the shapes displayed on the globe.

### Benefits
New feature that allow users to edit the shapes displayed on the globe.


markpet49:
@AkeluX @claudiaifrim Hi. I tried out this editor, the functionality is quite nice! In looking through the code, I noticed that the ShapeEditor is adding event listeners directly to the world window. 

Recently, the preferred method for capturing events was changed to adding a gestureListener to WorldWindow.worldWindowController (See the API of the class BasicWorldWindowController). This allows us to centralize UI event handling and provide consistent behaviour across all the devices we support (touch, mouse, mobile, etc.) If you need some pointers with respect to how to work with the controller, let me know.

examples/ShapeEditing.html
```diff
@@ -0,0 +1,53 @@
+<!DOCTYPE html>
```
AkeluX: Could we call this example ShapeEditor.html/.js so it matches with the feature name?

examples/ShapeEditing.js
```diff
@@ -0,0 +1,147 @@
+/*
+* Copyright 2015-2017 WorldWind Contributors
```
AkeluX: The header is misaligned and the copyright should be 2015 - 2018.

src/util/ShapeEditor.js
```diff
@@ -0,0 +1,1570 @@
+/*
+* Copyright 2015-2017 WorldWind Contributors
```
AkeluX: The header is misaligned and the copyright should be 2015 - 2018.

src/util/ShapeEditor.js
```diff
@@ -0,0 +1,1570 @@
...
+         * Drag on the shape's body moves the whole shape. Drag on a control point performs the action
+         * associated with that control point. The editor provides vertex insertion and removal for SurfacePolygon and
+         * SurfacePolyline. Shift-click when the cursor is over the shape inserts a control
+         * point at the cursor's position. Alt-click when the cursor is over a control point removes that control point.
+         * <p/>
+         * This editor supports all surface shapes except SurfaceImage.
+         * @param {WorldWindow} worldWindow The World Window to associate this shape editor controller with.
+         * @throws {ArgumentError} If the specified world window is null or undefined.
+         */
+        var ShapeEditor = function (worldWindow, shape) {
```
AkeluX: @pdavidc Should we call this class SurfaceShapeEditor for clarify?

pdavidc: Good question @AkeluX. We'll likely need editors for other shape types, like 3D polygons and the 3D airspace shapes we'll eventually develop. Do you think that this shape editor would be extended to support those shapes?
(Keep in mind, 3D shapes are displayed as surface shapes in 2D projection mode)

emxsys:
@claudiaifrim I've been experimenting with the ShapeEditor in one of my projects. I find I am unable to move a polyline -- I can only move individual vertices.  What's the technique for moving a polyline to a new location?

AkeluX:
@emxsys it's currently possible if you manage to click on the line, which is difficult with thin lines. At the moment, the only solution is to make the lines larger. We've actually discussed a few ideas to improve this with our TO today. We will be experimenting in the coming days.

emxsys:
@AkeluX Thanks for the feedback. I'll share two ideas for moving the polyline, just in case they're new:

1. Display a symbol that represents the polyline's reference position and use it for moving the shape. The symbol should be larger than a vertex and rendered behind vertices so that either it or a coincident vertex could be selected.
2. The current reference position for a polyline is the first vertex.  You could consider using the base of the "rotation lever" as the reference position for the polyline (and other shadow shapes).

PR #638, opened 2018-04-18
Enhanced the accuracy of ElevationModel's min/max elevations

pdavidc:
ElevationModel's min/max elevations interface is used indirectly by terrain and image tile rendering to determine level of detail. This change restores min/max accuracy to its pre-composition state, ensuring that terrain and image level of detail behaves as it has been.

- Corrected minElevation and maxElevation to return zero when the model has no coverages, or when all coverages are disabled
- Updated minAndMaxElevationsForSector to consult only the necessary coverages for a given sector
- Updated ElevationModel's unit tests to account for changes in the minAndMaxElevationsForSector interface and behavior
- Relates to epic #575

### DIFF
--- a/src/globe/ElevationCoverage.js
+++ b/src/globe/ElevationCoverage.js
@@ -97,13 +97,30 @@ define(['../error/ArgumentError',
         /**
          * Returns the minimum and maximum elevations within a specified sector.
          * @param {Sector} sector The sector for which to determine extreme elevations.
-         * @returns {Number[]} An array containing the minimum and maximum elevations within the specified sector,
-         * or null if the specified sector is outside this elevation model's coverage area or if the server data is not yet
-         * available.
-         * @throws {ArgumentError} If the specified sector is null or undefined.
+         * @param {Number[]} result An array in which to return the requested minimum and maximum elevations.
+         * @returns {Boolean} true if the coverage completely fills the sector with data, false otherwise.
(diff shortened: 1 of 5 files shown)


pdavidc:
I agree. However, indicating whether or not the min/max indicates whether the sector is filled is the whole intent of this change. Can you think of an alternate approach that might function just as well, while producing the intended result?

zglueck:
I'm missing nuances here which I'm sure are important, so I apologize ahead of time for dumb questions/suggestions!

Would it be better as two separate methods? e.g.
```
if (coverage.hasCompleteDataForSectpr(sector)) {
    coverage.minAndMaxElevationsForSector(sector, result)
}
```

pdavidc:
Yes, that would be a more concise design. However, it would incur the overhead of assembling and checking the coverage tiles for that sector twice. That's a fairly expensive operation.

I think the real problem here is that the elevation model system has taken on responsibilities that really don't belong to it. In this case, the elevation system is partially responsible for constructing terrain and image tile bounding boxes via minAndMaxElevationsForSector. If you look closely at how this method operates, you'll see behavioral that terrain and imagery depend on, but which don't really make sense for the elevation model. I have it in mind to address coupling between elevations and terrain/image tiles with a new epic.

In the meantime, we need an approach that's conceptually good enough, but performs well.

zglueck:
That sounds like the best solution! No additional comments from my end 👍

PR #640, opened 2018-04-23
Simplify the moveTo method in SurfaceCircle and SurfaceEllipse

claudiaifrim:
### Description of the Change
Refactoring of moveTo method for SurfaceCircle and SurfaceEllipse shapes.

Optimization for existing methods.

### Benefits
Optimization for existing methods.


pdavidc:
This is a good straightforward adjustment @claudiaifrim and @AkeluX. Thanks for taking care of this.

Issue #639, opened 2018-04-19
OGC Web Coverage Service (WCS) interface


pdavidc:
Child of #482 

Define an object model for interacting with an OGC Web Coverage Service (WCS). This model encapsulates the functionality of asynchronously discovering the service capabilities. Here are the use cases this functionality intends to support:

1. Given a service address, discover the coverages available.
2. Given a service address, construct WorldWind scene components for one or more coverages. 
3. Given a service address and a coverage name, construct a WorldWind scene component for that coverage.

zglueck:
This work was completed in pr #652 and work on the WebCoverageService.

PR #648, opened 2018-04-26
Add generic WCS Coverage object

zglueck:
### Description of the Change

This pr adds the initial implementation of the WcsCoverage object. The WcsCoverage object is a version agnostic simplistic representation of a WCS coverage that will be used to generate different WorldWind scene components.

### Applicable Issues

#639, #482


src/ogc/wcs/WcsCoverageDescriptions.js
```diff
@@ -59,15 +61,126 @@ define([
             this.assembleDocument();
         };
 
+        /**
```
pdavidc: Did you consider defining a `WcsCoverageDescription` object model that could encapsulate the responsibility of `getSector`, `getResolution`, and `getSupportedCrs`?

test/CustomMatchers.test.js
```diff
@@ -43,5 +43,23 @@ define([], function () {
         };
     };
 
+    CustomMatchers.toEqualSector = function (util, customEqualityTesters) {
```
pdavidc: Should this be a fuzzy comparison, or are the sectors we're checking in the unit test exact? Said another way, would this be better as `toBeSector`?

I have upcoming changes to CustomMatchers that adds `toBeVec3`, `toBeCloseToVec3` `toBePosition`, and `toBeCloseToPosition`.

src/ogc/wcs/WcsCoverage.js
```diff
@@ -18,28 +18,67 @@
...
+            this.coverageId = coverageId;
 
             /**
-             * A simple configuration object with the required parameters for ElevationCoverage.
-             * @type {Object}
+             * The WebCoverageService responsible for managing this Coveragea and Web Coverage Service.
+             * @type {WebCoverageService}
              */
-            this.elevationConfig = {};
+            this.webCoverageService = webCoverageService;
```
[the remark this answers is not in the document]
zglueck: `webCoverageService` is way too long an making some lines of the WcsUrlBuilder really long. I'll change to `service`, it doesn't lose meaning and is more concise.

src/ogc/wcs/WcsCoverageDescriptions.js
zglueck: No I hadn't thought about that approach. With the current approach, we've eliminated the need for a user to know the difference between versions and the index location. If those functions were refactored to singular WcsCoverageDescription(s) the notion of version or index would be required by the client unless the functionality was duplicated in parent WcsCoverageDescriptions.

What considerations am I missing in my thinking?

pdavidc: It seems to me there is some conceptual fuzziness in the WCS object model. In particular the relationship of WcsCoverageDescriptions and WcsCoverage. Here are two points I noticed that motivated me to comment:

• Extending WcsCoverage to facilitate future convenience methods requires changes to WcsCoverageDescriptions. That makes WcsCoverage look partially like a leaky abstraction, since for the purpose of extracting XML doc data WcsCoverage and WcsCoverageDescriptions are tightly coupled.
• Apps cannot easily query the WcsCoverage for information in the XML doc.

I can see why we'd want a WcsCoverage object model that's logically independent from the coverage summaries and coverage descriptions coming from the XML documents. The coverage descriptions for WCS 1.0 and 2.0 are quite different, but what if the common concept was represented by an object model in a manner similar to WcsCoverageDescriptions?

PR #649, opened 2018-04-27
TiledElevationCoverage config-based construction 

pdavidc:
Modified TiledElevationCoverage to enable construction via configuration object. This enables loose coupling between a WebCoverageService specifying the configuration, and the scene component that consumes the configuration (TiledElevationCoverage). This is the first of several changes that streamlines TiledElevationCoverage construction.

This change knowingly breaks backward compatibility. The class TiledElevationCoverage is new since 0.9.0. While we could argue that it should be possible to port 0.9.0 ElevationModel construction to TiledElevationCoverage, as the constructors currently appear very similar, they are fundamentally different. Additionally, the opportunity to improve the elevation coverage interface oughtweighs the benefit we might gain by making it easy to port old ElevationModel code to TiledElevationCoverage.

Relates to #486

### DIFF
--- a/src/globe/AsterV2ElevationCoverage.js
+++ b/src/globe/AsterV2ElevationCoverage.js
@@ -36,15 +36,21 @@ define([
          * @classdesc Provides elevations for Earth. Elevations are drawn from the NASA WorldWind elevation service.
          */
         var AsterV2ElevationCoverage = function () {
-            TiledElevationCoverage.call(this,
-                new Sector(-83.0001, 83.0001, -180, 180), new Location(45, 45), 11, "application/bil16",
-                "AsterV2Elevations256", 256, 256, 0.000277777777778);
+            TiledElevationCoverage.call(this, {
+                coverageSector: new Sector(-83.0001, 83.0001, -180, 180),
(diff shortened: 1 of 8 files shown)


markpet49:
@pdavidc Hi Dave. I like the change, it certainly improves the ease of configuring a coverage. Two questions:
1. Why is displayName left out of the config?
2. Would it be better to leave checks in place for required config properties in TiledElevationCoverage? Granted most of the parameters get checked by underlying code, but there are some that will result in cryptic errors if they aren't specified.

pdavidc:
@markpet49 my intent is to adjust the configuration parameters rapidly. This first iteration ports the existing constructor arguments, and adds those necessary for construction via WebCoverageService.

My intent going forward is that:
1. We'll remove unnecessary configuration parameters.
2. Configuration parameters are not a grab bag of anything an app might want to set at construction, but rather those parameters necessary to facilitate construction from services. I'm not sure what fits at the moment, so I left out anything unnecessary.
3. We'll add configuration parameter checking once we narrow down which parameters are required.

zglueck:
Everything looks good to me!

PR #650, opened 2018-04-27
Simplify TiledElevationCoverage tile management

pdavidc:
This change simplifies TiledElevationCoverage by removing usage of the unnecessary cachePath property. We could limit the scope of this change by assigning any string to cachePath and leaving TiledElevationCoverage's implementation unchanged. However, TiledElevationCoverage has significant unnecessary complexity that creates and tracks unique elevation tile keys based on the notion that cachePath is necessary. Elevation tiles and elevation images are are stored in an cache unique to each instance of TiledElevationCoverage (rather than a shared cache). Given that, there is no need to generate globally unique cache keys, which is the only purpose served by cachePath. It is enough to use Tile.tileKey, which can uniquely identify each tile in the coverage's tile cache and image cache.

This iteration removes the unnecessary ElevationTile class and replaces its usages with Tile. Additionally, this iteration removes the imagePath properties that derive from cachePath, replacing them with usage of Tile.tileKey or in-place generation of a tile key string (when a tile instance is not available). Given the size of this change, i have left cachePath intact, with the intent to remove it and its usages in the various TiledElevationCoverage implementations in the next iteration.

Relates to #486


pdavidc:
@markpet49 Thanks for catching this. TiledElevationCoverage.createTile is more unnecessary infrastructure we can remove. 

This function comes from the TileFactory interface, whose purpose is to enable code sharing amongst components that create tiles in the same manner, but need different types of tiles. It's used by Tile.prototype.subdivide and Tile.createTilesForLevel. That said, there is no need to have this interface on TiledElevationCoverage, as it operates on tiles in a fundamentally different way than other components (i.e. TiledImageLayer, Tessellator). I'll remove it from TiledElevationCoverage.

PR #652, opened 2018-04-30
Add elevation config property to WcsCoverage

zglueck:
### Description of the Change

The WcsCoverage object has been enhanced to provide a configuration object suitable for initializing TiledElevationCoverage. A WcsUrlBuilder was added to support the configuration requirements of TiledElevationCoverage. The new WcsUrlBuilder supports both WCS version 1.0.0 and 2.0.x and deprecates the WcsTileUrlBuilder.

### Applicable Issues
Relates to #486


pdavidc:
@zglueck I have incoming changes that vastly simplify TiledElevationCoverage configuration. Would it work for you if we merge those first, then simplify the WcsCoverage elevation configuration accordingly?

zglueck:
@pdavidc yes that works for me.

src/ogc/wcs/WcsCapabilities.js
```diff
@@ -61,6 +61,20 @@ define([
             this.assembleDocument();
         };
 
+        /**
+         * Returns the GetCoverage base url as detailed in the capabilities document
+         * @param coverageId the coverage id or name
+         */
+        WcsCapabilities.prototype.getCoverageBaseUrl = function () {
+            if (this.version === "1.0.0") {
+                return this.capability.request.getCoverage.get
```
pdavidc: Is there a typo on this line?

src/ogc/wcs/WcsCapabilities.js
```diff
@@ -61,6 +61,20 @@ define([
             this.assembleDocument();
         };
 
+        /**
+         * Returns the GetCoverage base url as detailed in the capabilities document
+         * @param coverageId the coverage id or name
```
pdavidc: Is this comment obsolete, or is getCoverageBaseUrl missing its coverageId argument?

src/ogc/wcs/WcsCoverage.js
```diff
@@ -72,13 +74,62 @@ define([
...
+            if (version === "1.0.0") {
+                // find the associated coverage description
+                for (var i = 0, len = this.service.coverageDescriptions.coverages.length; i < len; i++) {
+                    if (this.coverageId === this.service.coverageDescriptions.coverages[i].name) {
+                        availableFormats = this.service.coverageDescriptions.coverages[i].supportedFormats.formats;
+                        break;
+                    }
+                }
+            } else if (version === "2.0.1" || version === "2.0.0") {
+                availableFormats = this.service.capabilities.serviceMetadata.formatsSupported
```
pdavidc: This line is missing the terminating semicolon. Please double-check this hasn't occurred anywhere else.

src/ogc/wcs/WcsCoverage.js
```diff
@@ -72,13 +74,62 @@ define([
...
+            }
+
+            if (!availableFormats) {
+                return WcsCoverage.DEFAULT_FORMAT;
+            }
+
+            for (i = 0; i < WcsCoverage.PREFERRED_FORMATS.length; i++) {
```
pdavidc: Order n-squared loop. How about storing preferred formats as a map?

src/ogc/wcs/WcsUrlBuilder.js
```diff
@@ -0,0 +1,155 @@
...
+    function (ArgumentError,
+              Logger) {
+        "use strict";
+
+        /**
+         * Constructs key value pair (KVP) WCS GetCoverage urls.
+         * @param wcsCoverage the WcsCoverage for which the urls should be generated
+         * @constructor
+         */
+        var WcsUrlBuilder = function (wcsCoverage) {
```
pdavidc: Tight coupling between WcsCoverage and WcsUrlBuilder. What about providing the ID and service as construction arguments, since that's what's needed anyway?

src/ogc/wcs/WcsCapabilities.js
zglueck: Obsolete, actually, never actually needed.

src/ogc/wcs/WcsCapabilities.js
zglueck: All I'm seeing is the missing semicolon, was there a different error?

src/ogc/wcs/WcsCoverage.js
zglueck: Seems I missed a number of semicolons, thanks for the check

src/ogc/wcs/WcsCoverage.js
zglueck: @pdavidc will you provide a short demo of what you mean?

src/ogc/wcs/WcsCapabilities.js
pdavidc: I can't make sense of `getCoverage.get`

src/ogc/wcs/WcsCoverage.js
pdavidc: By map, I mean a data structure you can access by key, rather than by index. So the check to see if an available format matches a preferred format would be a constant time lookup.

src/ogc/wcs/WcsCapabilities.js
zglueck: The HTTP "GET" url as opposed to the "POST". Would it be better to change the names for 1.0.0 to look like `.methodGet` and `.methodPost`?

pdavidc: I see. Indeed the object names look like function names. Just do whatever you think is most consistent with WMS, and we can sort out any nuances like this later.

src/ogc/wcs/WcsCoverage.js
zglueck: I could use a map, but I'm not sure how to incorporate the fuzzy comparison.

pdavidc:
@zglueck It looks like you've addressed the points we discussed. Is there anything outstanding to do here?

zglueck:
@pdavidc the only thing unresolved was the property path for 1.0.0 GET url, e.g. `this.capability.request.getCoverage.get;`

pdavidc:
Does the functionality work? If so, let's come back to the WCS capabilities and describe coverage object models.

zglueck:
Yes it works, just a weird name.

PR #655, opened 2018-04-30
Updated the Standalone example elevation coverage

pdavidc:
EarthRestElevationCoverage, the coverage used by Standalone.html, was out of date after recent changes to TiledElevationCoverage. Updated this coverage implementation to correctly retrieve standalone elevation tiles.


pdavidc:
@NASAWorldWind/nasa-developers I've included you as reviewers to keep you in the loop on what WebWorldWind's standalone example looks like.

Let me know if you have any questions or concerns specific to this change.

PR #656, opened 2018-05-01
Feature/elevationsforgrid resolution

pdavidc:
Clarifies ElevationCoverage's elevationsForGrid interface by removing the unnecessary and potentially harmful targetResolution argument.

The function elevationsForGrid derived from the original ElevationModel, which used targetResolution to select a level in the model's tile pyramid. Since implementing composable elevations (#575), this function appears in both ElevationModel and ElevationCoverage, but serves a different purpose in each.

The argument list to elevationsForGrid need not be the same, and in fact should not be the same. ElevationModel needs targetResolution to let the caller influence which coverage is chosen to fulfil the request. ElevationCoverage does not need targetResolution. Allowing the caller to specify a resolution to ElevationCoverage is unnecessary and potentially dangerous. The grid itself provides an implicit resolution the coverage can use to select a level. Letting the caller override that resolution risks the caller unknowingly attempting more data than can can be downloaded and managed in memory.

Relates to #422

### DIFF
--- a/src/globe/ElevationCoverage.js
+++ b/src/globe/ElevationCoverage.js
@@ -139,14 +139,12 @@ define(['../error/ArgumentError',
          * @param {Sector} sector The sector for which to determine the elevations.
          * @param {Number} numLat The number of latitudinal sample locations within the sector.
          * @param {Number} numLon The number of longitudinal sample locations within the sector.
-         * @param {Number} targetResolution The desired elevation resolution, in degrees. (To compute degrees from
-         * meters, divide the number of meters by the globe's radius to obtain radians and convert the result to degrees.)
          * @param {Number[]} result An array in which to return the requested elevations.
          * @returns {Boolean} true if the result array was completely filled with elevation data, false otherwise.
          * @throws {ArgumentError} If the specified sector or result array is null or undefined, or if either of the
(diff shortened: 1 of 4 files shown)


pdavidc:
@NASAWorldWind/nasa-developers I've included you as reviewers to keep you in the loop on significant interface changes. Let me know if you have any questions or concerns about this change.

pdavidc:
@markpet49 This pull request eliminates the subtle multiple dependency on targetResolution introduced by having both ElevationModel and ElevationCoverage use the argument. After considered the implications of this decision, I came to the conclusion that attempting to use targetResolution in both classes was a mistake, for the reasons I outlined in this pull request's description.

I strongly recommend following this change with two others that I'm convinced will complete the #575:

1. Modify ElevationModel.elevationsForGrid to choose a starting coverage index by finding the coverage index with resolution nearest to targetResolution.

2. Modify Tessellator.regenerateTileGeometry and alignNeighborElevations to send a smaller targetResolution to Globe.elevationsForGrid. Perhaps we could provide a method on TerrainTile that encapsulates this behavior, since it's used in two places. I think we should apply a constant factor to Level.texelSize, as the behavior of that choice is easy to predict. Applications are likely to configure coverage sets that look nothing like our own.

PR #660, opened 2018-05-02
Add a reader for AAIGrid

strikerM:
### Description of the Change
Provides parsing capabilities for AAIGrid data.

Allows WebWorldWind to work with AAIGrid data.


src/formats/aaigrid/AAIGridReader.js
```diff
@@ -0,0 +1,249 @@
...
+                }
+            }
+        });
+
+        /**
+         * Returns the content of the AAIGrid as an Int16Array or Float32Array.
+         *
+         * @return {Int16Array|Float32Array} The content of the AAIGrid.
+         */
+        AAIGridReader.prototype.getImageData = function () {
```
AkeluX: How fitting is this method name for an AAIGrid?

PR #661, opened 2018-05-02
Replacement of Bing logo image credit for a ScreenImage

### LINKED ISSUE #659: Refactor Bing attribution as a screen image
Child of #489

The Bing layer's image attribution is a special case, and its requirements are unlike attributions coming from public services. Our relationship with the Bing team requires that the Bing logo must always be visible when the Bing layer is displayed, and there is no official text alternative to the Bing logo.

1. Replace the Bing layer's usage of image credits with a layer-owned screen image. Place it somewhere it does not conflict with existing screen shapes.

2. Automate retrieval of the Bing attribution URL from the web service, and remove the built-in Bing image.

3. Provide control over the Bing attribution screen location.

Beak-man:
### Description of the Change
Removed image credit usage from Bing's layers in order to replace it with a ScreenImage with the same contents.
Bing seems to be a special case among 3rd party imagery providers in that they require displaying their logo (instead of only string-based attributions) when their imagery is used. With this change we maintain compliance with Bing's TOS while setting the stage to simplify our general handling of attributions by relying only in text-based credits in a future step.

### Benefits
- Maintains compliance with Bing's TOS by displaying their logo when their tiles are displayed in the globe. 
- Enables us to handle credits independently of this logo in order to address credit cluttering and customization.
- Places the logo where screen real estate is available without operlapping other WebWW overlay elements. The logo accommodates itself when coordinates display is moved to the top of the canvas in small screens, utilizing the same resolution criteria as CoordinatesDisplayLayer.

### Potential Drawbacks
- Bing's logo is not retrieved from its web service. This was attempted but the URL provided by Bing's REST service does not allow CORS, plus it's hosted via insecure http. Another source that allowed hotlinking couldn't be found. This still uses our locally hosted Bing logo image file.

- Aesthetic choices on where to put the Bing logo. This can of course be changed after feedback from the team.

### Applicable Issues
Relates to #659.

### DIFF
--- a/src/layer/BingTiledImageLayer.js
+++ b/src/layer/BingTiledImageLayer.js
@@ -18,14 +18,22 @@
  */
 define([
         '../geom/Angle',
+        '../util/Color',
         '../geom/Location',
+        '../util/Offset',
+        '../shapes/ScreenImage',
         '../geom/Sector',
-        '../layer/MercatorTiledImageLayer'
+        '../layer/MercatorTiledImageLayer',
+        '../layer/ViewControlsLayer'
     ],
     function (Angle,
+              Color,
               Location,
+              Offset,
+              ScreenImage,
               Sector,
-              MercatorTiledImageLayer) {
+              MercatorTiledImageLayer,
+              ViewControlsLayer) {
         "use strict";
 
         /**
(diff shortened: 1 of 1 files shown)


Beak-man:
Bing logo was moved to the lower right corner.

src/layer/BingTiledImageLayer.js
```diff
@@ -80,6 +88,18 @@ define([
             return 256 << (levelNumber + 1);
         };
 
+        BingTiledImageLayer.prototype.createBingLogotype = function () {
+            // Locate Bing logo in the upper left corner
```
pdavidc: @Beak-man This comment and the one below are out-of-date.

src/layer/BingTiledImageLayer.js
```diff
@@ -80,6 +88,18 @@ define([
...
+            var offset = new Offset(WorldWind.OFFSET_FRACTION, 1, WorldWind.OFFSET_FRACTION, 0);
```
pdavidc: Let's place the Bing logo where it appears now. There's some amount of pixel inset, right?

Beak-man:
@pdavidc , @markpet49 , indeed filtering looks different. With screen credits, we apply opacity to the credits as a layer. In this implementation opacity is given to the screen image directly. Maybe that's where the difference stems from?

The image credit looks stretched by a single pixel compared to the logotype as implemented in this pull request. I couldn't replicate the exact alignment given by ScreenCreditController. The differences can be seen below (the screenshot with the OpenStreetMap attribution is the current implementation in develop).

binglogopositioning

Beak-man:
Positioning the logo 1px above the previous example. Still looks off. Probably due to differences in filtering?

binglogo2

PR #663, opened 2018-05-04
Added capability to modify Bing logo positioning and alignment

### LINKED ISSUE #659: Refactor Bing attribution as a screen image

Beak-man:
### Description of the Change
Added properties that allow modification of Bing's logo positioning and alignment on the screen and it's corresponding assignments in the layer's render loop.

### Benefits
- Allows the developer to modify the positioning of Bing's logo as desired.
- Maintains a pattern similarity with `ViewControlsLayer`.

### Potential Drawbacks
None. Modifying the logo's positioning was successfully tested.

### Applicable Issues
Closes #659

### DIFF
--- a/src/layer/BingTiledImageLayer.js
+++ b/src/layer/BingTiledImageLayer.js
@@ -62,15 +62,15 @@ define([
             this.attributionImage = WorldWind.configuration.baseUrl + "images/powered-by-bing.png";
 
             /**
-             * An {@link Offset} indicating where to place the Bing logo on the canvas.
+             * An {@link Offset} indicating where to place the Bing logo on the screen.
              * @type {Offset}
              * @default A value of (WorldWind.OFFSET_INSET_PIXELS, 5, WorldWind.OFFSET_PIXELS, 5) provides a
-             * 5px margin below and in the right side of the logo, placing it in the lower right corner of the canvas.
+             * 5px margin inset from the lower right corner of the screen.
              */
             this.logoPlacement = new Offset(WorldWind.OFFSET_INSET_PIXELS, 5, WorldWind.OFFSET_PIXELS, 5);
 
             /**
-             * An {@link Offset} indicating the alignment of the Bing logo relative to the logo placement.
+             * An {@link Offset} indicating the alignment of the Bing logo relative to the placement position.
              * @type {Offset}
              * @default Lower right corner of the logo.
              */
--- a/src/layer/BingTiledImageLayer.js
+++ b/src/layer/BingTiledImageLayer.js
@@ -61,11 +61,20 @@ define([
 
             this.attributionImage = WorldWind.configuration.baseUrl + "images/powered-by-bing.png";
 
-            // Bing logo placement on the screen. Lower right corner by default.
+            /**
+             * An {@link Offset} indicating where to place the Bing logo on the screen.
+             * @type {Offset}
+             * @default The lower right corner of the window.
+             */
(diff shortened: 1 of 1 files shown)


src/layer/BingTiledImageLayer.js
```diff
@@ -61,14 +61,26 @@ define([
 
             this.attributionImage = WorldWind.configuration.baseUrl + "images/powered-by-bing.png";
 
-            this.attribution = this.createBingLogotype();
+            // Bing logo placement on the screen. Lower right corner by default.
+            this.logoPlacement = new Offset(WorldWind.OFFSET_FRACTION, 1, WorldWind.OFFSET_FRACTION, 0);
```
pdavidc: The properties `logoPlacement` and `logoAlignment` need documentation. Otherwise this looks straightforward and well done.

src/layer/BingTiledImageLayer.js
```diff
@@ -61,14 +61,35 @@ define([
 
             this.attributionImage = WorldWind.configuration.baseUrl + "images/powered-by-bing.png";
 
-            this.attribution = this.createBingLogotype();
+            /**
+             * An {@link Offset} indicating where to place the Bing logo on the screen.
+             * @type {Offset}
+             * @default The lower right corner of the window.
+             */
+            this.logoPlacement = new Offset(WorldWind.OFFSET_FRACTION, 1, WorldWind.OFFSET_FRACTION, 0);
```
pdavidc: @Beak-man This documentation is perfect.

With fresh eyes this morning, it occurred to me that the placement and the alignment seem backwards. This concerns code clarity, not correctness. What you have works, but I think others will struggle to understand how the code relates to the logo's appearance. Try explaining what negative inset pixels means, and you'll see what I'm getting at.

The following configuration is more clear:
```
this.logoPlacement = new Offset(WorldWind.OFFSET_INSET_PIXELS, 5, WorldWind.OFFSET_INSET_PIXELS, 36);
this.logoAlignment = new Offset(WorldWind.OFFSET_FRACTION, 1, WorldWind.OFFSET_FRACTION, 0);
```

pdavidc:
@Beak-man Did you notice my latest comment on this pull request? It's a simple change, and I'm wondering why you haven't followed up yet.

src/layer/BingTiledImageLayer.js
pdavidc: That looks good @Beak-man. The two different logo units could be an easily overlooked subtlety, but at least the meaning is clear. Plus, there's no better way to achieve this.

You're right, we don't need to match ScreenCreditController's configuration exactly. The 5 pixel margin is easy to understand and adjust as needed. Additionally, the positions of text attributions and this Bing logo will likely change in the next iteration anyway.

Beak-man: It's done. There's another concern that I just noticed. The change of coordinates kind of reverses the meaning of "placement" and "alignment" if we go by the quantities. It now looks like `logoAlignment` is the one that locates the logo in the screen and `logoPlacement` provides the margins and alignment.

Issue #645, opened 2018-04-25
Attribution text clarity and decluttering


pdavidc:
Child of #489

Select the text style (interior and outline) that make attributions as readable as possible on the variety platforms WorldWind runs on, without being overbearing. 

Truncate text attributions to a single line with a maximum length, depending on the viewport dimensions. Suppress duplicate text attributions.

PR #681, opened 2018-05-07
Repositioned view controls, coordinates, and Bing logo overlays

Beak-man:
### Description of the Change
Every overlay present in the bottom of the canvas was repositioned with a 20px margin from it. This, in order to reposition screen credits below them as a single line.

### Applicable Issues
Related to #645

### DIFF
--- a/src/layer/BingTiledImageLayer.js
+++ b/src/layer/BingTiledImageLayer.js
@@ -67,7 +67,7 @@ define([
              * @default A value of (WorldWind.OFFSET_INSET_PIXELS, 5, WorldWind.OFFSET_PIXELS, 5) provides a
              * 5px margin inset from the lower right corner of the screen.
              */
-            this.logoPlacement = new Offset(WorldWind.OFFSET_INSET_PIXELS, 5, WorldWind.OFFSET_PIXELS, 5);
+            this.logoPlacement = new Offset(WorldWind.OFFSET_INSET_PIXELS, 5, WorldWind.OFFSET_PIXELS, 20);
 
             /**
              * An {@link Offset} indicating the alignment of the Bing logo relative to the placement position.
--- a/src/layer/CoordinatesDisplayLayer.js
+++ b/src/layer/CoordinatesDisplayLayer.js
@@ -146,9 +146,9 @@ define([
                 canvasWidth = dc.currentGlContext.canvas.clientWidth,
                 x, y, yUnitsScreen, yUnitsText, hideEyeAlt;
 
-            if (canvasWidth > 650) { // large canvas, align the text with bottom center
+            if (canvasWidth > 650) { // large canvas, align the text with bottom center, 20px margin from the bottom
                 x = (canvasWidth / 2) - 50;
-                y = 5;
+                y = 20;
                 yUnitsScreen = WorldWind.OFFSET_PIXELS;
                 yUnitsText = 0;
             } else if (canvasWidth > 400) { // medium canvas, align the text in the top left
--- a/src/layer/ViewControlsLayer.js
+++ b/src/layer/ViewControlsLayer.js
@@ -106,7 +106,7 @@ define([
             /**
              * An {@link Offset} indicating where to place the controls on the screen.
              * @type {Offset}
-             * @default The lower left corner of the window, with a 20px margin from the bottom side of the screen.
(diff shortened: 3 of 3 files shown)


pdavidc:
@Beak-man The change is perfectly straightforward. Given that it's aesthetic, I have two thoughts.

1. Is 20 pixels the minimum height we need for text credits?

2. Now that the vertical padding is greater, the horizontal padding for view controls and the Bing logo seems off.

@markpet49 It's worth your looking at this too.

Beak-man:
@pdavidc:
1. I think 20px will suffice. That being said, when the credits themselves are refactored (I'm on it) I may change the margin, depending on how it looks.

2. Maybe a 5px horizontal margin will look OK? The Bing logo has that in the other side. I'll try it.

src/layer/BingTiledImageLayer.js
```diff
@@ -67,7 +67,7 @@ define([
              * @default A value of (WorldWind.OFFSET_INSET_PIXELS, 5, WorldWind.OFFSET_PIXELS, 5) provides a
```
markpet49: This comment needs an update.

src/layer/BingTiledImageLayer.js
```diff
@@ -67,7 +67,7 @@ define([
...
              * 5px margin inset from the lower right corner of the screen.
              */
-            this.logoPlacement = new Offset(WorldWind.OFFSET_INSET_PIXELS, 5, WorldWind.OFFSET_PIXELS, 5);
+            this.logoPlacement = new Offset(WorldWind.OFFSET_INSET_PIXELS, 5, WorldWind.OFFSET_PIXELS, 20);
```
markpet49: Is it possible to make the y offset driven by the height of the text attributions rather than hard coded? In the default case where there are no attributions other than Bing, there is currently just blank unused space at the bottom.

markpet49:
I would hesitate to merge this change until the OSM credits are positioned properly. If you turn on Bing and OSM, the credits overlay each other. I wouldn't want the web examples to show this.

pdavidc:
@markpet49 The goal here is to make the two changes independently. I want this team to to move away from the larger pull requests I’ve been seeing to changes more narrow in scope. To accomplish this, screen elements may overlap temporarily while we’re in transition. Accepting temporary detours will be a part of very narrowly scoped changes, like those I’m encouraging @Beak-man to make.

Your point about dynamic height adjustment sounds right on. The constant vertical offer is odd when no attributions are present. 

@Beak-man how do you think we ought to proceed?

Beak-man:
Making the overlays' height dynamic sounds good, but then we could just leave the screen credits in their current position and either dynamically move the Bing logo on top of them, or just position it elsewhere.

I have another concern regarding drawing the credits in a single line. It looks good, even with every current attribution enabled in portrait mode:



But the approach required is not ideal. It requires foregoing the usage of multiple ScreenTexts in order to draw every credit in a single one. We lose the capability of assigning a hyperlink to each attribution this way, since the URL and click handler are attached to each ScreenText.

Another way to draw them in a single line would be to use separate ScreenTexts and then sizing each of them via their ActiveTextures in order to properly space them. Then we would retain hyperlinking... but of course this gets us again into the problem of reliably reading this property.

Maybe leaving the credits and layouts where they are, and repositioning the Bing logo is the best idea?

pdavidc:
@Beak-man You’re seem to be stuck on the idea that the texture dimensions must be known in order to layout individual items.

What about measuring the text with TextRenderer?

Beak-man:
@pdavidc , I see no other way when arranging the credits horizontally. Making a TextRenderer instance to size them will probably work. I suggest the following course of action then (each step being a PR):

- Addressing the bug outlined in #687 
- Making the appropriate changes to ScreenCreditController to lay out the credits horizontally, while also providing a flag so it can communicate whether or not there are screen credits to draw.
- With the aforementioned flag, enable dynamic positioning on the overlays related to this PR.

This PR could be closed in the meantime. With the steps outline aboce, there's no need to reposition these elements while the rest of the work is being done.

pdavidc:
@Beak-man this rough outline looks good. The implementation details of steps 2 and 3 need more thought before they’re worked on.

pdavidc:
Closing this issue with the expectation that a similar PR will be reopened.

PR #689, opened 2018-05-11
Addressed multiple Bing logos rendered on top of another

### LINKED ISSUE #687: Bing logo is rendered multiple times per each Bing layer
### Description
As an unintended effect of #661, each time a Bing layer is added to the globe, a new Bing logo is rendered as an overlay.

### Steps to Reproduce

1. Load a Bing layer (as in our examples, where Bing Aerial with Labels is shown on load).
2. Load another Bing layer.
3. See the Bing logo getting darker each time a Bing layer is added, and lightening each time a Bing layer is removed.

**Expected behavior:**
A single Bing logo is rendered no matter how many of Bing's layers are loaded into the globe.

**Actual behavior:**
Bing logos are rendered in top of another.

**Reproduces how often:**
100% of the time.

Beak-man:
### Description of the Change
Previous modification to `BingTiledImageLayer` added a Bing logo `ScreenImage` to it in order to provide imagery attributions  (#661). Inadvertently, this caused every subclass of `BingTiledImageLayer` to draw its own Bing logo when enabled, which resulted in multiple unnecessary logos being drawn in top of another.

After making sure this redrawing wasn't caused by an oversight in `BingTiledImageryLayer.doRender` (e.g. forcing the redraw to be coincident with `DrawContext.timestamp`), other methods were attempted to avoid this needless multiple instancing while maintaining previous architecture. After these attempts failed, the current proposal was drafted: As with every other 2D overlay, the logo is now added as its own layer and removed from `BingTiledImageryLayer` altogether. This way, the logo isn't copied to every subclass of this layer down the prototype chain, which was the cause of the bug.

In a similar manner as with `ScreenCreditController`, the new `BingLogoLayer` is attached to the `DrawContext` and added as an 'internal' layer of sorts by default to the `WorldWindow`. It only gets drawn when `BingTiledImageLayer` (or more appropriately, its subclasses) requests it.

Solves previous multiple Bing logo bug. 

### Benefits
Besides solving the bug, this maintains the pattern of providing every 2D overlay via a `Layer`. Maintains capability of logo placement customization.

### Potential Drawbacks
This layer is added to the `WorldWindow` in a similar manner to `ScreenCreditController`. I'm not sure if this is the proper pattern to follow. Unlike `ScreenCreditController`, this one is added into the `/src/layer` folder.

### Applicable Issues
Closes #687

### DIFF
--- a/src/WorldWind.js
+++ b/src/WorldWind.js
@@ -29,6 +29,7 @@ define([ // PLEASE KEEP ALL THIS IN ALPHABETICAL ORDER BY MODULE NAME (not direc
         './BasicWorldWindowController',
         './layer/BingAerialLayer',
         './layer/BingAerialWithLabelsLayer',
+        './layer/BingLogoLayer',
         './layer/BingRoadsLayer',
         './layer/BingWMSLayer',
         './layer/BMNGLandsatLayer',
@@ -281,6 +282,7 @@ define([ // PLEASE KEEP ALL THIS IN ALPHABETICAL ORDER BY MODULE NAME (not direc
               BasicWorldWindowController,
               BingAerialLayer,
               BingAerialWithLabelsLayer,
+              BingLogoLayer,
               BingRoadsLayer,
               BingWMSLayer,
               BMNGLandsatLayer,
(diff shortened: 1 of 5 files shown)


Beak-man:
@pdavidc I returned at my previous attempt at addressing this via JavaScript closures and finally got it working, without changing anything from the architecture. Please consider PR #690 . I'll close this PR.

PR #690, opened 2018-05-11
Avoided drawing of multiple Bing logos using static member

### LINKED ISSUE #687: Bing logo is rendered multiple times per each Bing layer

Beak-man:
### Description of the Change
Previous modification to BingTiledImageLayer added a Bing logo ScreenImage to it in order to provide imagery attributions (#661). Inadvertently, this caused every subclass of BingTiledImageLayer to draw its own Bing logo when enabled, which resulted in multiple unnecessary logos being drawn in top of another.

This change makes the `ScreenImage` object that produces the logo a static member of `BingTiledImageLayer`, so this object is shared among all subclasses of it, correcting the previous implementation which produced a copy of the logo for each subclass, hence producing the logo repetition.

In order to understand private and static members in JavaScript, I looked here:
https://crockford.com/javascript/private.html
https://www.quora.com/How-do-you-create-static-variables-in-JavaScript

### Benefits
Solves the aforementioned bug and keeps current architecture, unlike my first proposal (#689).

### Potential Drawbacks
- Not a pattern that I've seen elsewhere in Web WorldWind. 
- Pointing the `ScreenImage` to `this` may be unnecessary.

### Applicable Issues
Closes #687.

### DIFF
--- a/src/WorldWind.js
+++ b/src/WorldWind.js
@@ -949,6 +949,8 @@ define([ // PLEASE KEEP ALL THIS IN ALPHABETICAL ORDER BY MODULE NAME (not direc
          *     <li><code>baseUrl</code>: The URL of the directory containing the WorldWind Library and its resources.</li>
          *     <li><code>layerRetrievalQueueSize</code>: The number of concurrent tile requests allowed per layer. The default is 16.</li>
          *     <li><code>coverageRetrievalQueueSize</code>: The number of concurrent tile requests allowed per elevation coverage. The default is 16.</li>
+         *     <li><code>bingLogoPlacement</code>: An {@link Offset} to place a Bing logo attribution. The default is a 5px margin inset from the lower right corner of the screen.</li>
+         *     <li><code>bingLogoAlignment</code>: An {@link Offset} to align the Bing logo relative to its placement position. The default is the lower right corner of the logo.</li>
          * </ul>
          * @type {{gpuCacheSize: number}}
          */
@@ -957,20 +959,7 @@ define([ // PLEASE KEEP ALL THIS IN ALPHABETICAL ORDER BY MODULE NAME (not direc
             baseUrl: (WWUtil.worldwindlibLocation()) || (WWUtil.currentUrlSansFilePart() + '/../'),
             layerRetrievalQueueSize: 16,
             coverageRetrievalQueueSize: 16,
-
-            /**
-             * An {@link Offset} indicating where to place the Bing logo on the screen.
-             * @type {Offset}
-             * @default A value of (WorldWind.OFFSET_INSET_PIXELS, 5, WorldWind.OFFSET_PIXELS, 5) provides a
-             * 5px margin inset from the lower right corner of the screen.
-             */
             bingLogoPlacement: new Offset(WorldWind.OFFSET_INSET_PIXELS, 5, WorldWind.OFFSET_PIXELS, 5),
-
-            /**
-             * An {@link Offset} indicating the alignment of the Bing logo relative to the placement position.
-             * @type {Offset}
-             * @default Lower right corner of the logo.
-             */
             bingLogoAlignment: new Offset(WorldWind.OFFSET_FRACTION, 1, WorldWind.OFFSET_FRACTION, 0)
         };
 
--- a/src/WorldWind.js
+++ b/src/WorldWind.js
@@ -956,7 +956,22 @@ define([ // PLEASE KEEP ALL THIS IN ALPHABETICAL ORDER BY MODULE NAME (not direc
             gpuCacheSize: 250e6,
             baseUrl: (WWUtil.worldwindlibLocation()) || (WWUtil.currentUrlSansFilePart() + '/../'),
             layerRetrievalQueueSize: 16,
-            coverageRetrievalQueueSize: 16
+            coverageRetrievalQueueSize: 16,
+
+            /**
+             * An {@link Offset} indicating where to place the Bing logo on the screen.
+             * @type {Offset}
+             * @default A value of (WorldWind.OFFSET_INSET_PIXELS, 5, WorldWind.OFFSET_PIXELS, 5) provides a
+             * 5px margin inset from the lower right corner of the screen.
+             */
+            bingLogoPlacement: new Offset(WorldWind.OFFSET_INSET_PIXELS, 5, WorldWind.OFFSET_PIXELS, 5),
+
+            /**
+             * An {@link Offset} indicating the alignment of the Bing logo relative to the placement position.
+             * @type {Offset}
+             * @default Lower right corner of the logo.
+             */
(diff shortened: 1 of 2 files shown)


Beak-man:
Regardless of what's decided about this proposal, I'd like to share some findings about JS access specifiers and static members, which may be of use if we choose to make use of these features in the future, given our current tooling (particularly our AMD-based module definition).

I took a while to come up with this small change since most code online shows the following as an example of a static member:

````javascript
var myClass = function ( ) {
  this.member1 = "foo"; // Public member in constructor
  var myVariable = "secret"; // Private to the constructor's scope
}
myClass.prototype.member2 = "bar"; // Public member outside constructor via prototype
myClass.member3 = "baz" // Static member
````
So what I attempted at first was to make the `ScreenImage` logo a static member by declaring it outside the constructor. Due to scope, I naturally can't point to `this` outside of the constructor if the member is not attached to the `prototype`. So I attempted the following:
````javascript

var BingTiledImageLayer = function (displayName) {
    // ...
};

BingTiledImageLayer.bingLogo = new ScreenImage(
    new Offset(WorldWind.OFFSET_INSET_PIXELS, 5, WorldWind.OFFSET_PIXELS, 5),
    WorldWind.configuration.baseUrl + "images/powered-by-bing.png");

BingTiledImageLayer.prototype = Object.create(MercatorTiledImageLayer.prototype);
// ...
````
But RequireJS won't bind the `WorldWind` module required for the `ScreenImage` to the `bingLogo` member (and I suspect that it won't obtain any module at all in this manner, WebStorm warns that there's a missing AMD module dependency).

This is specific for our current tooling and it took me a while to understand that I needed to declare the static member inside the constructor.

In [Crockford's link](https://crockford.com/javascript/private.html) there's an explanation about creating a privileged method as well.

**Comment by pdavidc** [comment]
@Beak-man This still isn’t the right approach. You’re trying to solve a frame association problem with JavaScript class mechanics.

**Comment by Beak-man** [comment]
@pdavidc OK, closing this as well.

**Comment by Beak-man** [comment]
@pdavidc , as we talked earlier today, the logo last frame time is now the static member instead. The logo itself is back to how it was (copied between the subclasses). 

This works, but I'm not sure if it's worthwhile to keep the logo as static in order to avoid needless copying of it between subclass instances, or if maintaining the usual pattern of non-static members should take precedence.

**Comment by Beak-man** [comment]
@pdavidc : 
1. Understood. Changed it to 'logoLastFrameTime'.
2. In that case, the conceptual association that takes precedence should be the class and its subclasses, since the logo has to appear on every Bing layer. So maintaining the logo as non-static seems fitting.

**Comment by pdavidc** [comment]
Regarding 2, my take was the opposite. The logo's association with the Bing layer class indicates it should be static.

**Comment by Beak-man** [comment]
Both `attribution` (the logo ScreenImage) and `logoLastFrameTime` are set as static members now. Everything is in the constructor due to the need of ScreenImage to be inside a closure, as explained before.

Just out of curiosity, I tested setting the logo as static and the last frame time as non-static. It works OK. Would that approach be conceptually valid?

**Comment by pdavidc** [comment]
@Beak-man While the Bing logo does indeed appears only once, this code does not do what it intends to. I'm surprised you've submitted it for review in the state it's in. Please take a more careful look at this code.

**Comment by Beak-man** [comment]
@pdavidc You're right, I mistakenly left outdated references to `this.logoLastFrameTime` from the test I mentioned above. It's corrected now.

**Comment by Beak-man** [comment]
@pdavidc , I added the appropriate documentation for new members of `WorldWind.configuration`.

As we talked earlier, alongside `bingLogoPlacement` and `bingLogoAlignment`, I'm proposing adding `bingLogoUrl`. However, for it to work it needs accessors in order to refer to `this.baseUrl` for its default value. 

Please consider if adding this boilerplate to the configuration object is acceptable since it makes it somewhat less readable. The advantage of having it is that it provides the developer with a simple, consistent interface to change everything about the Bing logo. 

I hotlinked an alternative logo from [here](https://www.microsoft.com/en-us/maps/mobile-brand-guidelines) and I tested it by adding to our BasicExample the following: 
````javascript
WorldWind.configuration.bingLogoUrl = "https://img-prod-cms-rt-microsoft-com.akamaized.net/cms/api/am/imageFileData/RW9wQe?ver=a6ea";
````
(Funny that Bing allows hotlinking on this logo and not on the one referred in their REST service metadata). It works OK and I tested it with another locally sourced image as well.

Another change to consider is that since the Bing logo URL is now intended to be modified from the WorldWind configuration object, `BingTiledImageLayer` is slightly simplified. I removed the previous check for `if(!BingTiledImageLayer.logoImage)` from `BingTiledImageLayer.renderLogo`. Now the function always overwrites the `BingTiledImageLayer.logoImage` with a ScreenImage constructed with whatever is assigned to `WorldWind.configuration.bingLogoUrl`. 

I added some comments explaining why `BingTiledImageLayer.logoImage` assignment to a ScreenImage is not done in its declaration, which is kept as internal.

File: src/WorldWind.js [hunk 1]
```diff
@@ -949,14 +949,20 @@ define([ // PLEASE KEEP ALL THIS IN ALPHABETICAL ORDER BY MODULE NAME (not direc
...
          * </ul>
          * @type {{gpuCacheSize: number}}
          */
         WorldWind.configuration = {
             gpuCacheSize: 250e6,
             baseUrl: (WWUtil.worldwindlibLocation()) || (WWUtil.currentUrlSansFilePart() + '/../'),
             layerRetrievalQueueSize: 16,
-            coverageRetrievalQueueSize: 16
+            coverageRetrievalQueueSize: 16,
+            bingLogoUrl: (WWUtil.worldwindlibLocation()) || (WWUtil.currentUrlSansFilePart() + '/../') + "images/powered-by-bing.png",
```
pdavidc: Remove this. We don't intentionally duplicate code for such a minor benefit.

src/layer/BingTiledImageLayer.js
```diff
@@ -103,12 +86,25 @@ define([
             this.topLevelTiles.push(this.createTile(null, this.levels.firstLevel(), 1, 1));
         };
 
+        BingTiledImageLayer.prototype.renderLogo = function (dc) {
+            // Assigning ScreenImage to logoImage in here due to circular dependency of
```
pdavidc: Lazy object creation is a common pattern in WorldWind. This comment doesn't add anything.

PR #691, opened 2018-05-14
Add the notion of measured locations

balhar-jakub:
**Note:** Filling out this template is required. Any pull request that does not include enough information to be reviewed in a timely manner may be closed at the
maintainer's discretion.

### Description of the Change

Support for the location containing the measure. 

This notion is present in GeoJSON, Heatmaps and other formats and as such makes sense to be part of the core. 

### Applicable Issues

Closes #691


pdavidc:
@balhar-jakub @AkeluX This addition of MeasuredLocation is straightforward and well done. All the considerations of inheriting from Location are accounted for.

There are a few minor grammatical typos in the documentation, but otherwise this looks good to go. @balhar-jakub @AkeluX would you be willing do another proof reading of the documentation text?

balhar-jakub:
I did another proofreading. 
@AkeluX  Could you also take a look whether I missed something?

AkeluX:
@balhar-jakub @pdavidc, while updating the API doc for this feature, I looked at the methods in Location and had a couple of thoughts I wanted to share with you.

1) Does it make sense to define a MeasuredLocation.ZERO?
2) If it does, wouldn't it make sense to have MeasuredLocation.fromRadians too?
3) Moreover, do we expect the other static methods in Location to work with MeasuredLocations, meaning that the measure is actually carried over to the result?

Let me know what you think.

pdavidc:
@Akelux thank you for catching this.

1. I’m not sure what task MeasuredLocation.ZERO helps with. If neither you nor Jakub can think of one, perhaps it’s best to leave it out.

2. MeasuredLocation construction ought to be complementary to standard Locations. I think fromRadians makes sense even if we omit ZERO.

3. Location’s static methods ought to work. Otherwise MeasuredLocation is crippled. Many operate on lat/lon and may operate correctly now. Are there any that need to be extended to accommodate MeasuredLocation?

AkeluX:
1. I don't. @balhar-jakub?

2. That is simple to achieve. Will do.

3. They do work fine for lat/lon if passed `MeasuredLocation` objects as parameters.
My question is namely about `greatCircleLocation`, `rhumbLocation` and `linearLocation`, which transforms a single location and put the results in the `result` parameter. If both the original location and the result are `MeasuredLocation` objects, do we expect the result object to contain the measure from the original location? This of course doesn't work at the moment.
I think that it doesn't make sense to consider the other static methods because they work on two locations and it would be impossible to decide which measure to put in the result object.

balhar-jakub:
1) There is no reason. It isn't used internally and so the potential effectivity gain is zero and it is probably better not to provide it for public usage. 
Updated in the commit.

2) It really does make sense. 
Updated in the commit. 

3) I believe that wherever possible we should also provide the static methods on the MeasuredLocation.  It removes the necessity to load the Location directly for the usage of the methods once you already loaded the MeasuredLocation.

The following methods should be directly supported on the MeasuredLocation as the fact that the location is measured doesn't change anything for them. The implementation should be done by delegation the work to the implementations on the Location:
greatCircleAzimuth
greatCircleDistance
greatCircleLocation
rhumbAzimuth
rhumbDistance
rhumbLocation
linearAzimuth
linearDistance
linearLocation
locationsCrossDateLine
intersectionWithMeridian
meridianIntersection

All the following methods contain multiple locations and returns Location as result. As such it is unclear what measure could be used. I believe that we should provide these methods on the MeasuredLocation as well. For the implementation on the MeasuredLocation I would like to add another parameter which is the measure which should be used for result/s 
If none is provided than the MeasuredLcoation will be returned with measure set to null (This is something I am realy unsure about)
interpolateAlongPath
interpolateGreatCircle
interpolateRhumb
interpolateLinear
greatCircleArcExtremeLocations
greatCircleArcExtremeForTwoLocations
greatCircleExtremeLocationsUsingAzimuth

What are your thoughts?

AkeluX:
These methods being static and already available from Location, I would not "copy" them in MeasuredLocation, even if this is only pointing to their counterpart in Location. This would make maintenance a lot harder for benefits I fail to see at this point.

If anything, I would make sure that the ones in Location work as expected with MeasuredLocation too. In that respect, the only doubt I have is the one mentioned in my previous comment under point 3.

balhar-jakub:
Well the only real benefit is clarity. 
When the methods are static on MeasuredLocation it is obvious that they work with MeasuredLocation. For methods which aren't part of the MeasuredLocation they shouldn't be expected to work correctly with MeasuredLocation. 

If this isn't from your point of view, strong enough argument, then the solution for the first set of method is obviously to ignore them as they will work as they are. In case of 
rhumbLocation
linearLocation
greatCircleLocation
update them to work correctly with measure as there is just one location in the beginning and in the result. The measure should be copied. 

The issue is with following list of methods:
interpolateAlongPath
interpolateGreatCircle
interpolateRhumb
interpolateLinear
greatCircleArcExtremeLocations
greatCircleArcExtremeForTwoLocations
greatCircleExtremeLocationsUsingAzimuth
These methods accept multiple locations but return just one. Here it is absolutely unobvious what should be the measure in result. I explained the possibilities from my point of view in previous comment 

From the comment (All the following methods contain multiple locations and returns Location as result. As such it is unclear what measure could be used. I believe that we should provide these methods on the MeasuredLocation as well. For the implementation on the MeasuredLocation I would like to add another parameter which is the measure which should be used for result/s
If none is provided than the MeasuredLcoation will be returned with measure set to null (This is something I am realy unsure about))

AkeluX:
@pdavidc if it's fine with you, I propose to merge the current state. I've created an issue for tracking the remaining open points, see #739, which doesn't affect the currently implemented parts.

pdavidc:
@AkeluX @balhar-jakub The current state of this change looks good to me. I concur that the minimum implementation is best for now.

Issue #484, opened 2018-02-23
WCS compatible elevation tiling


pdavidc:
Closes #482

Adjust ElevationModel's tiling to achieve compatibility with non-global WCS coverages. 

ElevationModel adopted the tiling approach used by WMS layers, where top level tiles span the globe. This happens to work with global WCS coverages, but breaks when a WCS coverage does not span the globe. WCS responds to tile requests exceeding a coverage bounds in a manner unlike WMS: the response is stretched to fill the image, and the assumption of response values containing no-data (like WMS transparency) fails.

The most straightforward and reliable mechanism for broad WCS compatibility is to adjust ElevationModel's approach to tiling. Use a tiling scheme that fits to the WCS coverage bounds, rather than the LevelSet notion of global tiles. See WorldWindAndroid's TiledElevationCoverage for an example implementation.

PR #693, opened 2018-05-16
Add TileMatrix/TileMatrixSet

zglueck:
### Description of the Change

This pr ports WWA's TileMatrix and TileMatrixSet classes to WebWW for future support of regional elevational coverages.

### Applicable Issues

See #484


src/geom/TileMatrixSet.js
```diff
@@ -0,0 +1,143 @@
...
+            return new TileMatrixSet(sector, tileMatrices);
+        };
+
+        /**
+         * Determines the index of the TileMatrix with the closest resolution to the provided value.
+         * @param degreesPerPixel the target resolution
+         * @returns {number} the index of the TileMatrix within this TileMatrixSet's entries array
+         */
+        TileMatrixSet.prototype.indexOfMatrixNearest = function (degreesPerPixel) {
+            if (degreesPerPixel < 0) {
```
markpet49: This check could be '<=0', could it not?

src/geom/TileMatrix.js
```diff
@@ -0,0 +1,144 @@
...
+
+            /**
+             * The geographic coverage area of this TileMatrix.
+             */
+            this.sector = sector;
+
+            /**
+             * The index position of this TileMatrix in the TileMatrixSet.
+             */
+            this.ordinal = ordinal;
```
markpet49: Could we use a more descriptive variable name here? LevelSet/Level uses levelNumber, but I'm not sure that's semantically correct here.

src/geom/TileMatrix.js
```diff
@@ -0,0 +1,144 @@
...
+             * The number of tiles in the vertical or y direction.
+             */
+            this.matrixHeight = matrixHeight;
+        };
+
+        /**
+         * Calculate the resolution of this TileMatrix in degrees per pixel
+         * @returns {number} degress per pixel resolution
+         */
+        TileMatrix.prototype.degreesPerPixel = function () {
```
markpet49: Could this be a property populated at construction time similar to Level.texelSize? It might make sense if the underlying properties (sector, tileHeight, etc.) aren't too dynamic.

src/geom/TileMatrix.js
```diff
@@ -0,0 +1,144 @@
...
+         * @constructor
+         */
+        var TileMatrix = function (sector, ordinal, matrixWidth, matrixHeight, tileWidth, tileHeight) {
+            if (!sector) {
+                throw new ArgumentError(
+                    Logger.logMessage(Logger.LEVEL_SEVERE, "TileMatrix", "constructor",
+                        "missingSector"));
+            }
+
+            if (ordinal < 0) {
```
markpet49: These numeric tests pass if the variable is undefined which, unfortunately, is a likely error case if you forget to specify all the parameters.
You could add some unit tests to validate this e.g.
        it("test", function () {
            expect(function () {
                TileMatrixSet.fromTilePyramid(sector);
            }).toThrow();
        });

zglueck: I started with a more explicit check (e.g. typeof) but didn't see that pattern in other geom objects so I reverted to the bounds checking.

pdavidc: Numeric argument checks are intentionally anything other than range checks. Early on we took the position that it isn't worth the trouble it takes to meaningfully protect the caller from themselves in this JavaScript. It's an adjustment I've seen every seasoned Java developer go through.

src/geom/TileMatrix.js
pdavidc: This came from Android, but I don't recall whether or not it was necessary.

@zglueck Is ordinal used anywhere?

zglueck: @pdavidc it is used in the Tile key for TiledElevationCoverage

pdavidc: Let's remove TileMatrix.ordinal. It was a shim that doesn't really belong. Additionally, I'll need to change how elevation tile keys are generated to accommodate the change in language.

src/geom/TileMatrixSet.js
zglueck: @markpet49 On second thought, is there a value in allowing zero, say for requesting the highest resolution case?

zglueck: For my edification, what about when `0` is a legal value? What should a check look like?

pdavidc: Nevertheless, @markpet49 highlights an important point. The argument checking pattern requires a judgement call, and different authors have been making different decisions.

As Mark suggested, how about we use the standard check when zero is not a legal value:
`if (!degreesPerPixel)`

markpet49:
@zglueck As usual good discussion, the latest looks good to me.

PR #701, opened 2018-05-21
Add surface shape performance test

### LINKED ISSUE #694: Surface shape performance functional test
Child of #561 

Create a functional test for benchmarking surface shape performance

- Determine key attributes to test and report
- Clearly define and communicate results

zglueck:
### Description of the Change

This PR adds a functional test which measures SurfaceShape performance. Both static and moving surface shapes are tested using an automated navigation technique.

### Applicable Issues

Closes #694

### DIFF
--- a/functionalTest/SurfaceShapes.html
+++ b/functionalTest/SurfaceShapes.html
@@ -16,23 +16,12 @@
             width: 100%;
             height: 100vh;
         }
-
         #buttons {
-            position: absolute;
-            top: 30px;
-            left: 30px;
-            z-index: 10000;
+            padding-top: 10px;
         }
     </style>
 </head>
 <body>
-<div id="buttons">
-    <div class="btn-group" role="group" aria-label="Basic example">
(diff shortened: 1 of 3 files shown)


zglueck:
@pdavidc I've addressed 2, 3, and 4; however I need guidance with 1.

With the `functionalTest` directory designated as a test directory WebStorm's server function makes the `content-type` header equal `text/html`. The javascript files are then prevented from executing because:
```
because its MIME type ('text/html') is not executable, and strict MIME type checking is enabled.
```

Using an alternative server (e.g. python) allows the functional test to run, but WebStorm's built-in seems to be changing the content type header on my machine.

Is there a workaround for the content type in WebStorm?

pdavidc:
@zglueck The content-type error is not the root of the problem. Registering the functionalTest folder as a test root affects the configuration of WebStorm's built-in web server. The paths to resources in src no longer work as expected.

PR #709, opened 2018-05-23
Commented boilerplate code for examples (from Paths to WMTS).

Beak-man:
### Description of the Change
Added comments to every example boilerplate code regarding layer creation and addition to the WorldWindow, as well as explaining the Logger configuration. Some examples were lacking documentation, and these blocks of code are common to all our examples (with some slight variations on amount of layers being used, etc).

This PR relates only to our examples from Paths to WMTS (in alphabetical order). The same work from Annotations to ParseUrlArguments will be addressed in a separate PR.

No logic was modified. The comments were only added or repositioned.

### Applicable Issues
Relates to #539

### DIFF
--- a/examples/StarField.js
+++ b/examples/StarField.js
@@ -28,7 +28,7 @@ requirejs([
         // Create the WorldWindow.
         var wwd = new WorldWind.WorldWindow('canvasOne');
 
-        // Create layers.
+        // Create imagery layers.
         var BMNGLayer = new WorldWind.BMNGLayer();
         var starFieldLayer = new WorldWind.StarFieldLayer();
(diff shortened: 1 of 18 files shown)


examples/StarField.js
```diff
@@ -22,17 +22,20 @@ requirejs([
               LayerManager) {
         'use strict';
 
+        // Tell WorldWind to log only warnings.
         WorldWind.Logger.setLoggingLevel(WorldWind.Logger.LEVEL_WARNING);
 
+        // Create the WorldWindow.
         var wwd = new WorldWind.WorldWindow('canvasOne');
 
+        // Create layers.
```
eirizarry: Should be "Create imagery layers".

PR #710, opened 2018-05-23
Commented boilerplate code for examples (from Annotations to ParseURLArgs).

eirizarry:
### Description of the Change

Added comments to every example for logging configuration, worldwindow, and layers to give examples a consistent boilerplate.

### Applicable Issues


examples/BlueMarbleTimeSeries.js
```diff
@@ -23,10 +23,13 @@ requirejs(['./WorldWindShim',
               LayerManager) {
         "use strict";
 
+        // Tell WorldWind to log only warnings.
         WorldWind.Logger.setLoggingLevel(WorldWind.Logger.LEVEL_WARNING);
 
+        // Create the WorldWindow.
         var wwd = new WorldWind.WorldWindow("canvasOne");
 
+        // Create and add layers to the WorldWindow.
```
Beak-man: This example focuses in a single animated imagery layer, so layer creation and addition is handled differently. I suggest not straight replacing and copying the comments present in the other examples, but complementing the current comments with our usual boilerplate.

examples/BlueMarbleTimeSeries.js
```diff
@@ -40,7 +43,7 @@ requirejs(['./WorldWindShim',
         // which is initiated below.
         wwd.addLayer(blueMarbleTimeSeries);
 
-        // Create a compass and view controls.
+        // WorldWind UI layers.
```
Beak-man: Maybe "Create Worldwind UI layers"?

examples/Collada.js
```diff
@@ -14,19 +14,25 @@
  * limitations under the License.
  */
 
-require(['./WorldWindShim', './LayerManager'], function (WorldWind, LayerManager) {
+require(['./WorldWindShim', './LayerManager'],
```
Beak-man: Good catch. If this is going to be modified, we should maintain our usual indentation, e.g:
````javascript
require(['./WorldWindShim',
         './LayerManager'],
     function (WorldWind,
               LayerManager) {
````

File: examples/GeoJSONExporter.js [hunk 4]
```diff
@@ -42,6 +47,9 @@ requirejs(['./WorldWindShim',
             wwd.addLayer(layers[l].layer);
         }
 
+        // Create a layer manager for controlling layer visibility.
```
Beak-man: Leaving the LayerManager creation at the end works OK. Since it's something prevalent in all our examples, and its functionality is separated from the feature that the examples themselves refer to, we should avoid having it in the middle of the example (which is something that we should address soon in a lot of them).

examples/Measurements.js
```diff
@@ -31,13 +31,18 @@ requirejs([
...
 
+        // Create the WorldWinow.
         var wwd = new WorldWind.WorldWindow('canvasOne');
 
+        // Create and add layers to the WorldWindow.
+        // Imagery layers.
         var BNMGLayer = new WorldWind.BMNGLayer();
-        var CoordinatesDisplayLayer = new WorldWind.CoordinatesDisplayLayer(wwd);
         var pathLayer = new WorldWind.RenderableLayer('Path');
+        // WorldWindow UI layers.
```
Beak-man: In this case, the comment refers to a single layer, so it should be written in singular.

pdavidc:
FYI @eirizarry : issues and pull requests need labels and milestones assigned.

PR #713, opened 2018-05-23
Add atmosphere layer to black background to all examples

eirizarry:
### Description of the Change

Last step in changing Atmosphere Layer to a Sun Simulation Example. Added the atmosphere layer with a black css-styled background. 

### Applicable Issues

### DIFF
--- a/examples/ScreenText.js
+++ b/examples/ScreenText.js
@@ -33,9 +33,9 @@ requirejs(['./WorldWindShim',
             // Imagery layers.
             {layer: new WorldWind.BMNGLayer(), enabled: true},
             {layer: new WorldWind.BMNGLandsatLayer(), enabled: false},
+            {layer: new WorldWind.BingAerialWithLabelsLayer(null), enabled: true},
             {layer: new WorldWind.AtmosphereLayer(), enabled: true},
             // WorldWindow UI layers.
-            {layer: new WorldWind.BingAerialWithLabelsLayer(null), enabled: true},
             {layer: new WorldWind.CompassLayer(), enabled: true},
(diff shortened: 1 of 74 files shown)


Beak-man:
Overall, the change looks good. I attempted adding the atmosphere layer after other application imagery layers are created (e.g. SurfaceImage) and it doesn't seem to make any difference: The atmosphere doesn't change the color of other elements over the globe, besides the base imagery.

There are two examples where the atmosphere produces glitches. As I mentioned in Slack, in MultiWindow, where the outer layer of the atmosphere doesn't appear in the second and third globes:

screenshot 2018-05-24 11 36 07

And in the ScreenText example, where only the outer layer is shown, with no "inner" shading over the globe:

screentextatmosphere

@pdavidc , even with these 2 glitches, I consider we should just go ahead with the merge while documenting them. What do you think?

examples/ScreenText.js
```diff
@@ -33,6 +33,7 @@ requirejs(['./WorldWindShim',
             // Imagery layers.
             {layer: new WorldWind.BMNGLayer(), enabled: true},
             {layer: new WorldWind.BMNGLandsatLayer(), enabled: false},
+            {layer: new WorldWind.AtmosphereLayer(), enabled: true},
             // WorldWindow UI layers.
             {layer: new WorldWind.BingAerialWithLabelsLayer(null), enabled: true},
```
Beak-man: The BingAerialWithLabelsLayer is mistakenly added after the AtmosphereLayer, which produces the latter to not show its inner shading. BingAerialWitlLabelsLayer should go before the atmosphere and after BMNGLandsatLayer.

PR #715, opened 2018-05-25
Improved documentation for ParsingURLArguments and Custom Placemarks Examples

eirizarry:
### Description of the Change

Boilerplate code for the logging, WorldWindow, and layers were set at the top of both examples. 
In Custom Placemarks, we refactored the code to read more logically. i.e. setting the geographic position, creating a placemark, setting common attributes. 

### Applicable Issues


Beak-man:
@pdavidc , you're very right. @eirizarry pointed this out when we were pair programming. I was also terribly confused by those PlacemarkAttributes groupings at my internship and came to accept them as a fact of life.

We rearranged that code in order to remove unnecessary PlacemarkAttributes object creations and overwrites. It's now more in line to what we have in Annotation: We first define the attributes and then we feed them to the shape constructor.

PR #716, opened 2018-05-25
Documentation improvement of four examples

Beak-man:
### Description of the Change
Improved readability and comments of the following examples
- Collada
- DeepPicking
- GeographicMeshes
- GeographicText

Regarding the last example, a really long array declaration containing the data of U.S. mountains was moved away from the example code and added as an AMD module (In the file GeographicTextPeaks.js). 

The example functionality remains the same, but now the reader doesn't have to scroll through a long object array declaration in order to get to what the example is about.

### Benefits
Increased readability and improved documentation.

### Potential Drawbacks
We're adding a new module with the U.S. peaks data. It may be considered unnecessary. Adding it as a JSON file was also considered, but this approach avoids adding the necessary code to retrieve an external file.

### Applicable Issues
Relates to #539


PJHogan:
Whoa! Endia is approving Pull Requests!

Issue #730, opened 2018-06-05
GeoTiff example freezes temporarily before loading file


Beak-man:
### Description

The globe becomes unresponsive seconds after page load and right before the application loads the GeoTiff file (shown here by attempting to drag the globe up until the GeoTiff loads):

![geotiff](https://user-images.githubusercontent.com/7622894/40980138-62fef25c-689d-11e8-8739-8aef6115d832.gif)

If the example is already on cache, the canvas simply shows no globe until the GeoTiff loads. Then it shows the globe and takes the camera to the GeoTiff imagery.

### Steps to Reproduce

1. Load GeoTiffExample
2. Continuously drag the globe
3. See the dragging becoming unresponsive
5. The globe becomes responsive again after the GeoTiff loads

**Expected behavior:**
No unresponsiveness.

**Actual behavior:**
Freezing and unresponsiveness before the GeoTiff loads.

**Reproduces how often:**
100% of the time, although differently depending on if the example is cached or not.

### Operating System and Version
Windows 10 under Chrome and Firefox.

[What operating system and version are you using?]

### Additional Information

Any additional information, configuration or data that might be necessary to reproduce the issue.

Issue #724, opened 2018-06-01
SurfaceImage Example picking inconsistencies


Beak-man:
### Description
The SurfaceImage example has as a secondary feature the showcasing of the handling of picking events over the images. The idea is that whenever a SurfaceImage is picked, the console conveys that the event happened through a message.

But the message happens when the cursor is hovered anywhere over the globe, regardless of if there's a SurfaceImage in the location or not:

![surfaceimagepicking](https://user-images.githubusercontent.com/7622894/40862636-c38f4ae2-65b2-11e8-9a58-392f600d2789.gif)

### Steps to Reproduce

1. Load SurfaceImage example.
2. Open the browser's developer console.
3. Hover over where the SurfaceImages are present and over where they aren't.
4. See console messages building up.

**Expected behavior:**
The console message "Surface image picked" only appears when hovering over surface images.

**Actual behavior:**
The console message is triggered while hovering over anywhere in the globe.

**Reproduces how often:**
100%

### Operating System and Version
Windows 10, Chrome and Firefox.

Beak-man:
As part of #539, @eirizarry and myself were considering removing the picking code from the example altogether, but we decided to document the issue as to not end up just hiding a potential bug by removing the feature from the example.

Issue #723, opened 2018-06-01
GoToLocation example picking feature does not work


Beak-man:
### Description
The description of the GoToLocation example states:
"Illustrates how to use a GoToAnimator to smoothly move the view to a new location. This example listens for terrain picks and moves the view to the location picked."

The code implies an attempt of handling the picking and tapping events in order to fulfill this, but the view is never changed regardless of where the user picks or taps in the globe.

### Steps to Reproduce

1. Load GoToLocation.
2. Pick somewhere over the globe.

**Expected behavior:**
The view relocates to the where the user picks or taps.

**Actual behavior:**
The view remains unchanged.

**Reproduces how often:**
100%

### Operating System and Version
Windows 10, Chrome and Firefox. Also attempted to use the dev tools' simulated mobile view to emulate tapping.

PR #731, opened 2018-06-05
Readability and documentation enhancements in five examples

Beak-man:
### Description of the Change
Five examples were reviewed and enhanced in terms of their documentation and readability:
- GeoTiffExample
- GoToLocation
- Paths
- RefreshImagery
- SurfaceImage

### Summary
- GeoTiffExample and KmlExample were renamed to 'GeoTiff' and 'KML'. Their test configurations and corresponding HTML files were modified accordingly.
- GeoTiffExample (now GeoTiff) had its code rearranged. We're proposing removing the named callback function with an anonymous one in order to maintain the narrative-like flow. We see a similar approach in the Collada example. If the callback was too complex, or if the example made use of multiple callbacks (e.g. like in GeoJSON), then maintaining the named callback would probably be a better fit.
- GoToLocation only had a type coercive operator corrected.
- Paths and RefreshImagery had some comments correctly capitalized.
- SurfaceImage had its SurfaceImages order of creation rearranged so that instead of going from static image -> dynamic image -> static filtered image, it now creates the two static images first and then the dynamic one, progressively increasing code complexity. We're also proposing removing the picking code from this example, since the picking feature is already showcased in other examples and it needlessly complicates readability. There's a caveat on doing this, however, since we found a bug while testing (see below). It may be of interest to keep the picking feature, at least until the bug is solved.

- Additionally, KmlExample was renamed as KML in order to align its naming with GeoTiff. Thoroughly reviewing this example is not part of this PR.

### Known bugs
- GeoTiffExample. The globe freezes temporarily while the file is being loaded (#730).
- GoToLocation. Picking features are not working (#723).
- SurfaceImage. The confirmation that the images are being picked is wrongly notified while hovering everywhere over the globe. Not apparent until the dev console is open (#724).

### Applicable issues
Relates to #539.


Beak-man:
@AkeluX , I'm tagging you just to notify you about the KML example name change. We're merging this PR, but if there's any concern from your team about it, we can address it afterwards.

AkeluX:
Thanks for letting us know. No concerns at this point.

Issue #721, opened 2018-05-30
Region picking example does not work on mobile


eirizarry:
### Description
Found this bug in the Region Picking Example
### Steps to Reproduce

Run Region Picking example on mobile/tablet.

**Expected behavior:** 

Able to pick through by tapping on screen

**Actual behavior:** 

Nothing happens

**Reproduces how often:** 

100%

### Operating System and Version

n/a

### Additional Information

n/a

pdavidc:
@eirizarry FYI: the title suggests to me that WorldWind's picking subsystem is broken, which as we know isn't the case.

PR #735, opened 2018-06-12
Improvements on clarity and documentation for 5 examples

Beak-man:
### Description of the Change
Improved code clarity and documentation in the following examples:
- GeoJSONExporter
- Measurements
- Paths
- PickAllShapesInRegion
- SurfaceShapes

A rundown of the changes:

**GeoJSONExporter**
The shared attributes were placed at the beginning of the example. The example previously featured every surface shape available to Web WorldWind. In order to make it more brief, some surface shapes were removed. @pdavidc , please consider if this change is fitting, or if all surface shapes have to be featured in this example.

**Measurements**
Rearranged DOM bindings placement in the source file. Added comments.

**Path**
Replaced outdated TODO comment. The work on the example can be considered completed now.

**PickAllShapesInRegion**
Rearranged the layer creation so it comes right after the creation and addition of our "boilerplate" layers. Added comments.

**SurfaceShapes**
Moved the common attributes to the beginning, and renamed every shape variable so instead of simply rewriting the same "shape" var, each one has a name that relates to its shape type. Moved the polyline to the end, since it uses a different set of attributes. Added some comments.

### Known bugs
PickAllShapesInRegion doesn't work in mobile environments. Documented in #721

### Applicable Issues
Relates to #539


Beak-man:
@AkeluX , if there's any concern on reducing the scope of GeoJSONExporter (we propose reducing the amount of SurfaceShapes featured in the example), we'll be glad to address it.

AkeluX:
@Beak-man, no concerns. Quite the contrary, the overall work done on the examples lately has been really great in my opinion.

Beak-man:
Thanks @AkeluX!

Issue #734, opened 2018-06-08
Change our animation mechanism in the Examples


Beak-man:
### Description
We're currently using animation features in 3 examples:
- SunSimulation
- TestMovingSurfaceShapes
- StarField

The first two utilize `Window.setInterval` as a mechanism to update their animations, while the latter uses `WorldWindow.redrawCallbacks`. 

Neither mechanism is fitting for the task at hand (animation). SetInterval leaves the developer with the burden of defining a rate at which the animation will happen, which will always be arbitrary and will never synchronize the animation with the monitor's refresh rate.

`WorldWindow.redrawCallbacks` used in the last example on the list, happens to be very opaque to the developer, being difficult to understand without diving deep into the innards of Web WorldWind.

After touching base with @pdavidc , it has been decided to change both of these approaches to the [requestAnimationFrame](https://www.w3.org/TR/html51/webappapis.html#animation-frames) browser mechanism. This provides a standarized way to do animations without the opaqueness of WorldWindow.redrawCallbacks (which internally also uses requestAnimationFrame). Compared to setInterval, requestAnimationFrame maintains synchronization with the monitor's refresh rate. It also has some nice bonuses, like suspending the animation work if the browser's window is not on focus.

We will be refactoring the examples listed above to utilize requestAnimationFrame, as part of the ongoing work on improving our Examples and documentation (#539).

Issue #348, opened 2017-11-17
StarField edge is visible when globe is tilted


strikerM:
### Description

StarField edge is visible when globe is tilted to the horizon (tilt = 90)

### Steps to Reproduce

1. Open StarField example
2. Tilt the globe to 90 deg
3. Zoom in/out until the StarField edge is visible

**Expected behavior:** StarField edge should not be visible

**Reproduces how often:** 100%

### Operating System and Version

macOS High Sierra 10.13.1 / Firefox 57.0 (64-bit)
iOS 11.1.1 / Safari 11.0

### Additional Information

![ss_star_filed_edge](https://user-images.githubusercontent.com/14248863/32947262-c0d58b12-cba3-11e7-8b1e-0684aad5717f.png)

PR #753, opened 2018-06-18
StarField example refactored as "Night Sky Animation"

### LINKED ISSUE #726: Run StarField example time passage simulation automatically
### Description

The StarField example features a passage of time simulation similar to the one seen in the Sun Simulation example (formerly "Atmosphere layer" example).

The addition of a checkbox to start the simulation introduces an unnecessary requirement of interaction from the user, who may be confused if he's expecting something to happen right after the example loads.

Making the simulation to run automatically on app load will remove this potential confusion, while simplifying the example code and HTML document.

This change was previously done as part of the refactor of the Atmosphere Layer example when it became the Sun Simulation example.

Related to #539

Beak-man:
### Description of the Change
Changes the StarField example animation mechanism. Changes the focus of the overall example.

**In regards to the animation mechanism:**
We're now using the standard `requestAnimationFrame` instead of `WorldWindow.redrawCallbacks` as an animation mechanism. 
As mentioned by @pdavidc , while `redrawCallbacks` internally uses `requestAnimationFrame` to do the animation, using that feature from WorldWindow in this example promotes confusion. By resorting to the standard, we align the example with usual JavaScript practices for animation tasks.

The animation is time-based, as it adjusts accordingly to framerate. Regardless of the performance of the device where it's executed on, a day in the simulation should last 8 real time seconds. This time lapse was chosen arbitrarily, I just think it's speedy enough to immediately convey that something is changing in the scene, without being unpleasantly quick.

We're also removing the previously required user interaction to initiate the animation, simplifying the code and avoiding potential confusion, since it wasn't obvious that the user had to click on somewhere to initiate the animation.

**In regards to changing the focus of the example**
The idea of the original StarField example was showcasing the features of its related layer, in a similar manner as it is with the Atmosphere layer example, which also features a similar day/night cycle animation.

Since #713 we added the atmosphere layer to most of the examples, rendering the Atmosphere layer example almost redundant, except for its animation features — which are replicated in the StarField example anyway. I'm proposing focusing this example on the animation itself (and the `time` properties of both the `StarFieldLayer` and `AtmosphereLayer`) for three reasons:
 - We have no examples that focus on animation specifically, and this change of focus may be useful for developers looking into animating stuff in Web WorldWind, at least until we do an example that focuses completely on animation.
- We can always do another simple example that, for instance, showcases the StarField layer without animation, if we deem it necessary. That would be a more appropriate "StarField Layer Example", since the animation feature is not part of the StarFieldLayer proper. (It has to be noted that we haven't added the StarFieldLayer to all the examples due to #348).
- We can delete the Sun Simulation example, since its day/night cycle feature is showcased here (and with the proper animation mechanism, in that example we were using `SetInterval`), and the Atmosphere layer without animation is now shown in almost all the examples.

### Potential Drawbacks
@pdavidc , please consider if my approach on adjusting the animation to framerate is appropriate. It may prove confusing to a reader since this particular animation focuses on Earth's time passage (simulated time) and animation adjustment to the frame time (real time). I tried my best on distinguishing both "times" in the example, so your opinion on my clarity will be most appreciated.

More importantly, I'm not sure if I properly conveyed how I achieved the adjustment to frame time. The `timeToAdvance = frameTime * 10800` line is pretty much using a magic number, and while I try to explain where it came from, it's in this line where I'm referring to the two conceptually different sets of "times" simultaneously (the simulated time of day in `timeToAdvance` and the real time in `frameTime`). This distinction is not obvious and the explanation on the `10800` may prove unsatisfactory. What do you think? We can always forego the adjustment to framerate in order to simplify the example.

### Applicable Issues
Relates to #539 and #734
Closes #726


Beak-man:
@eirizarry and me arranged to change the name of the example to "Night sky animation". Will update the PR to reflect that.

examples/NightSkyAnimation.js
```diff
@@ -0,0 +1,86 @@
...
+        wwd.addLayer(atmosphereLayer);
+
+        // Set a date property for the StarField and Atmosphere layers to the current date and time.
+        // This enables the Atmosphere layer to show a night side (and dusk/dawn effects in Earth's terminator).
+        // The StarField layer positions its stars according to this date.
+        var now = new Date();
+        starFieldLayer.time = now;
+        atmosphereLayer.time = now;
+
+        // Animate the starry sky as well as the globe's day/night cycle.
```
pdavidc: Comment without code. Needs to be moved or removed. Probably best above line 82.

PR #756, opened 2018-06-20
Add an inertial effect when a fling gesture is detected

AkeluX:
A fling is a high-velocity drag. In our case, the idea is to continue the effect of the drag for a certain period of time with a nice effect.

The current implementation is a first attempt and is based on the following assumptions:
- A fling is a drag ending at 100 or more pixels per second in X or Y.
- The animation lasts for 1500ms.
- The easing function is a sinusoidal out.
- The latitude and longitude are the only elements that need to be kept track of for detecting external changes to the navigator.
- The only gesture that needs to terminate an existing animation is another drag.

The `minVelocity` and `animationDuration` could become properties of the `BasicWorldWindowController`. The other assumptions need to be verified through further analysis and testing.

Comments and suggestions are welcome. Tested only on Chrome so far.


pdavidc:
@AkeluX This looks like a very straightforward and solid approach. The changes are isolated, the easing function is simple and effective, and the use of the browser's animation frame API is perfect. Nicely done!

A few things came to mind:

1. Let's encapsulate the logic that detects the fling gesture and reports its velocity by either...
a. Implementing FlingRecognizer (and allowing it to recognize simultaneously with the controller's drag recognizer).
b. Adding gesture velocity properties to DragRecognizer.
2. I'd expect all gestures to terminate the fling. Navigation close to the surface may be difficult otherwise. To accomplish that, we'll likely need to compare the navigator's model-view matrix in the same manner you're doing with look-at location.
3. Applying the fling translation linearly in geographic coordinates is okay in 3D, but it doesn't work well in 2D. The controller has separate drag functions for 3D and 2D. Fling would need to follow that pattern.
4. Can we guarantee that the delta converges to 0.0 when elapsed is 1.0? With past navigator animation effects, we've had subtle bugs where this assumption breaks, causing the animation to continue indefinitely. It may be more robust to explicitly stop the animation when its duration has elapsed.

AkeluX:
@pdavidc thank you for the feedback. Below, I've extended on your first point. Let me know what you think.

1. a) If we go with this option, would we consider a fling recognizer that works for both mouse and touch events? This seems to contradict the current pattern of having recognizers specific to the type of input, e.g. click/tap or drag/pan.
I think the biggest problem with this option is not technical, but is to find two names for the new recognizers ;)

1. b) It's unfortunately not only drag, but also pan, that has to be considered in this case. This would be rather simple to implement.
Pan and drag don't have a common base class that could contain this velocity property besides the base gesture recognizer class. It seems that this base class contains almost, but not all, properties describing the resulting gestures. What is the criterion for putting a new property there or in the recognizer subclass?

2. I've tested only with the whole globe in view so far and it works fine, e.g. if you zoom or tilt while the animation is on-going. It actually feels pretty natural. I will try close to the ground and report back.

3. Yes, that's a point I forgot to mention. The current implementation is only for the 3D globe. That was on purpose because I think it's the most beneficial use case. My proposition is to handle the 2D case later on.

4. Yes, it will be more robust to stop the animation based on the elapsed percentage. I will change this.

AkeluX:
Points 2 and 4 from @pdavidc were addressed and a proposition for point 1 was added.

The base class `GestureRecognizer` handles the properties that are relevant for multiple, but not necessarily all, concrete recognizers. In particular, this class gives a common name to concepts that differ from touch and mouse events. My proposition is thus to add properties to this class for `translationVelocityX` and `translationVelocityY` like there are already for the actual translation delta.

An important point is that the relevant velocity for the fling is not the average velocity since the start of the gesture, but the velocity in the last moments of the gesture.
For example, one could press a mouse button and move round. Then, without releasing this mouse button, the mouse stays in place for a long time (at gesture scale), before finally making a sudden movement while its button is released. Although the average translation velocity over the gesture time might be very slow, in the last moments, it should be enough to trigger a fling. This is what is implemented now.

AkeluX:
As agreed with @pdavidc, I will implement a FlingRecognizer that handles both touch and mouse events.

AkeluX:
Added the `FlingRecognizer`. This new recognizer and the changes to `GestureRecognizer` could be provided as a separate PR if needed.

@pdavidc please let me know if you have any comments.

AkeluX:
As discussed with @pdavidc, check the behaviour with touches, e.g. put one finger down, then another one and lift the first one, etc.

Issue #737, opened 2018-06-14
KML classes are not defined in WorldWind global


emxsys:
I have a need to override the default attributes used in `KmlPlacemark` and `KmlStyle` but I am unable to do so because these classes are not declared in the `WorldWind` global.

    TypeError: WorldWind.KmlPlacemark is undefined.

balhar-jakub:
I believe that the KML Classes should be part of the API, as it is possible and often really beneficial to overwrite parts of them and provide custom implementations. I wrote about how to do it here:
https://medium.com/@jakub_43881/kml-support-in-the-web-world-wind-23d94a9e5e58 but without them being part of the public API the instructions won't work. 

If we don't want to clutter the WorldWind space we can expose them on the KmlFile as a static properties? 

@AkeluX  @pdavidc  Any thoughts on your side?

AkeluX:
We had a similar discussion about GeoJSON a while ago. The conclusion was roughly that one of the advantages of Web WorldWind is that it can be extended by the applications, e.g. tailoring of an existing shape beyond its attributes. Consequently, even internal classes should be exposed via the WorldWind namespace. @pdavidc do you agree?

pdavidc:
@AkeluX Yes, I agree with this conclusion. Additionally, exposing all classes via the a single mechanism, the WorldWind global, is vital.

pdavidc:
Closed via #759

PR #759, opened 2018-06-22
Expose internal KML classes in the WorldWind namespace

balhar-jakub:
### Description of the Change

As per #737 I prepared a PR with the exposure of internal KML classes on the WorldWind Object. 

To simplify customization of KML for developers.

### Benefits

Every user can decide how will the KML Elements look and provide implementations of the elements that can handle more complex behavior. 

### Potential Drawbacks

To some extent internals become part of the public API. 

### Applicable Issues

#737


AkeluX:
That's going to be a mess, but I believe that all that these classes should all be prefixed with `Kml` in order to avoid cluttering the namespace with very generic names like Update and Change without any context.

PR #764, opened 2018-06-28
Refactored ScreenText Example

Beak-man:
### Description of the Change

An alignment bug slipped through the review of PR #752. The bug was caused due to an overlooked reference at the TextAttributes.offset of the Center ScreenText. 

Moreover, this refactoring proposes a change of focus about showcasing the usage of TextAttributes.offsets. Instead of focusing on modifying the offsets after ScreenText creation, now it shows an arguably more application-oriented use-case: How TextAttributes.offset is used for text alignment. It also further clarifies the distinction between the former offsets and the screen offsets.

There are now two ScreenTexts positioned at the center of the screen, and their final position is defined by their alignment. 

The code is still pretty much a "single-page" read, accomplishing our previous brevity target for this example.

### Applicable Issues
Relates to #539

### DIFF
--- a/examples/ScreenText.js
+++ b/examples/ScreenText.js
@@ -52,7 +52,7 @@ requirejs(['./WorldWindShim',
         // Create a Renderable Layer to hold the ScreenTexts.
         var textLayer = new WorldWind.RenderableLayer("Screen Text");
 
-        //Create left ScreenText
+        // Create left ScreenText
         screenOffset = new WorldWind.Offset(WorldWind.OFFSET_FRACTION, 0, WorldWind.OFFSET_FRACTION, 0.5);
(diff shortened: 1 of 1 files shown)


Beak-man:
At one time @eirizarry and I considered the default font size to be too small. What do you think?
I agree on removing the Font color customization. That, we just dragged it from how the example previously was.

screen shot 2018-06-28 at 11 31 18

pdavidc:
@Beak-man @eirizarry The screen shot above looks good enough to me. By the way, I like how this example now serves to teach developers how to apply the two offsets to achieve common effects.

eirizarry:
Nice, @Beak-man! And the example reads so much simpler now 👍

PR #765, opened 2018-06-28
Repaired BlueMarbleTimeSeries Examples

Beak-man:
### Description of the Change
This change corrects previous caching behavior that pervaded the BMNGRestLayer from ever loading in the example, while also making it simpler to read.

### Benefits
Corrects previous faulty behavior, probably overlooked by me in #383 due to using browser cached contents. [Edit: No, it was due to changes done on layer network requests in #533]. Greatly simplifies the example.

### Potential Drawbacks
Moved the UI layer creation to the top. This is uncharacteristic compared with other examples, but having these layers created in the middle broke the flow of the narration of the example. Due to technical necessities, the AtmosphereLayer creation is still in the same location, though.

### Applicable Issues
Relates to #539


Beak-man:
@pdavidc I was just playing it safe, since I think I reviewed the last PR regarding this example!
In the "Description" section I wasn't referring to browser's cache, but how the Example didn't finish pre-populating the sub-layers.

pdavidc:
@Beak-man Thanks for clarifying.

To be clear, this bug certainly wasn't your doing! PR #533 broke this example by changing the conditions of layer network requests. The changes in #533 are correct, good, and should have been introduced much earlier in WebWorldWind's lifecycle. Nevertheless, it broke the assumption that this example made in terms of how prepopulating works.

pdavidc:
@eirizarry You have the last word on whether you see any changes needed.

Beak-man:
Ah, thanks for the peace of mind and the explanation, @pdavidc.

Issue #631, opened 2018-04-12
Collada shape model limitations


angelozinzi:
Hi, I am trying to figure out if World Wind Web could be used as a novel visualization system for MATISSE (http://openplanetary.co/blog/tools/matisse.html).
Even if it should be easily adapted for larger objects (such planets and moons), for small, irregular objects (comets, asteroid) a 3D shape model needs to be used.
I therefore tried to use a shape model in collada format for the comet 67P (target of the ESA Rosetta mission) and I found out that large shape models (i.e., larger than 100000 facets) are not correctly visualized (only in World Wind Web, whereas they are well displayed with normal collada viewers).
To be precise, the object is visualized up to facet number 100000 (more or less, I don't know exactly the value, but I noted differences when using files with more than 100000 facets).

Is there a limiting keyword somewhere in the World Wind Web code? Could this be easily disabled if not needed?

Furthermore, if I try to put the shape model in the centre of the reference system (the origin, the center of the Earth) the shape disappears. I can use every altitude (negative too) larger than -6378137. With this exact value I have no object displayed.

Even in this case I would like to know if it is possibile to easily overcome this problem or it is reccomended not to use World Wind Web to visualize such objects.

Thanks,
Angelo

strikerM:
Hi,

Can you provide a minimal test case with the large model?

Objects at the center of the Earth are probably culled by the gpu.
Try to use a bigger altitude, for example -6000000 or -5000000

angelozinzi:
Hi, in attachment 3 test cases:
 - 67p_lo is a shape model with 31456 vertices and 62908 polygons
 - 67p_mid has 833538 vertices and 1667072 polygons
 - pumpkin has 144016 vertices and 288028 polygons


![67p_lo](https://user-images.githubusercontent.com/14196558/38725413-8d94841a-3f07-11e8-9773-30748b2de399.png)
![67p_mid](https://user-images.githubusercontent.com/14196558/38725414-8db02210-3f07-11e8-8cf5-ba76b8ab1550.png)
![pumpkin](https://user-images.githubusercontent.com/14196558/38725415-8dcbf602-3f07-11e8-9370-0dd2446ec405.png)

As you can see, 67p_lo is correctly visualized (on the left the world wind web visualization, on the right the collada visualization with a python viewer); the pumpkin shape is visualized without the petiole from the body; 67p_mid is very poorly visualized (I think only a small part of a lobe of the shape is shown).

[...]

PR #766, opened 2018-06-29
Improved Collada rendering

strikerM:
### Description of the Change
**1. Flatten the node tree structure and merge all geometry in one or two buffers.**
This creates at most 2 gpu buffers per model rather than 4 per mesh.
One buffer is created for attributes(positions, normals and texture coordinates) and one buffer for indices.
Other than positions, the rest are optional and this method does not require that meshes have normals, texture coordinates or indices.
This includes mixing all sort of combinations, some meshes may only have positions, some have all the attributes and so on.

**2. Frustum culling.**
Frustum culling is done by the position of the model, rather than the bounding box.
This has the advantage that it’s faster, but it will cull the model at the edge of the frustum if the position is outside the frustum but some part of the model is still visible.
In my opinion this is not noticeable and outweighs having to compute the bbox and test against it.

**3. Reduced object initialisation during rendering.**
Vec3, Matrix and Color instances where created during rendering and then discarded.
They are now initialized once in the constructor and then reused.

**4. Added support for large models.**
Support for large models was added by using the ```OES_element_index_uint extension```.
This extension has good support and is strait forward to implement.
In case the model requires the extension and the extension is not available, an error is logged to the console and rendering proceeds as normal. The model will most likely not be rendered correctly, some parts will be missing.

### Benefits
Faster rendering of collada models.
Support for large models.

### Potential Drawbacks
None.

### Applicable Issues
#631

### DIFF
--- a/src/formats/collada/ColladaScene.js
+++ b/src/formats/collada/ColladaScene.js
@@ -899,27 +899,22 @@ define([
 
         // Internal. Intentionally not documented.
         ColladaScene.prototype.initializeUintExtension = function (dc) {
-            if (!this._uintExtCacheKey) {
-                this._uintExtCacheKey = dc.gpuResourceCache.generateCacheKey();
+            var ext = dc.gpuResourceCache.resourceForKey(ColladaScene._uintExtCacheKey);
+
+            if (ext) {
+                this._hasUintExt = true;
             }
-
-            var ext = dc.gpuResourceCache.resourceForKey(this._uintExtCacheKey);
-            var extensionInCache = !!ext;
-
-            if (extensionInCache === this._hasUintExt) {
-                return;
-            }
-
-            this._hasUintExt = false;
-
-            if (!ext) {
+            else {
+                this._hasUintExt = false;
                 ext = dc.currentGlContext.getExtension('OES_element_index_uint');
                 if (ext) {
-                    dc.gpuResourceCache.putResource(this._uintExtCacheKey, ext, 1);
+                    dc.gpuResourceCache.putResource(ColladaScene._uintExtCacheKey, ext, 1);
                     this._hasUintExt = true;
                 }
             }
         };
+
+        ColladaScene._uintExtCacheKey = 'uintExtCacheKey';
 
         return ColladaScene;
 
--- a/src/formats/collada/ColladaScene.js
+++ b/src/formats/collada/ColladaScene.js
@@ -50,7 +50,6 @@ define([
          * info needed to render the scene.
          */
         var ColladaScene = function (position, sceneData) {
-
             if (!position) {
                 throw new ArgumentError(
                     Logger.logMessage(Logger.LEVEL_SEVERE, "ColladaScene", "constructor", "missingPosition"));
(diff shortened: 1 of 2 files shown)


AkeluX:
@strikerM could you describe briefly how you tested these changes?

src/formats/collada/ColladaScene.js
```diff
@@ -667,223 +816,106 @@ define([
...
+            var ext = dc.gpuResourceCache.resourceForKey(ColladaScene._uintExtCacheKey);
+
+            if (ext) {
+                this._hasUintExt = true;
+            }
+            else {
+                this._hasUintExt = false;
+                ext = dc.currentGlContext.getExtension('OES_element_index_uint');
+                if (ext) {
+                    dc.gpuResourceCache.putResource(ColladaScene._uintExtCacheKey, ext, 1);
```
AkeluX: Couldn't the extension be saved in a member variable of the class? Instead of the _hasUintExt, we could then check whether this new member variable is null or not for example.

What is the advantage of using the GPU resource cache?

strikerM: It will not work in a multi window setup.
GPU resources can not be shared between contexts, so only the first window will render the model, the rest will throw a webGL error.

However this approach has a flaw that the extension may be evicted from the cache and then we will have a conflict of state, where the instance says it’s true and the gpuResourceCache says it’s false.

I propose we don’t keep any information about the extension on the instance and just use the gpuResourceCache for now.

Alternatively we may look into a better approach about managing gl extensions.
For example, we can use a dedicated class or simply cache them on the drawContext and use a get method ```dc.getExtension('OES_element_index_uint')```

strikerM:
For functional testing I used the models in the apps folder and some other models that I have locally.
In total I used 18 models with varying geometry attributes combinations, including two large models, and one with no indices.
I tested that picking still works in combination with other renderables such as placemarks and that the all the models render in a multi window setup.

For benchmarks I used the profiler tools from Chrome.
The new rendering method is a couple of times faster.
The biggest improvement comes form merging the buffers.

src/formats/collada/ColladaScene.js
AkeluX: Dave and I discussed this on Tuesday. The right place to store this is indeed as part of the DrawContext. It also detects context loses and can clear the cache accordingly. Could you propose the required changes to the DrawContext in a separate pull request? Thanks.

AkeluX:
@strikerM #769 was merged. Could you adapt this PR to it? Thanks.

strikerM:
@AkeluX It doesn't work with large models. Typed arrays don't have a ```name``` property so the condition always evaluates to ```false```.
Why the change from ```instanceof```?

AkeluX:
`instanceof` is one of the slowest ways of checking for an object type in JavaScript I believe. 

According to 
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Uint16Array, there is a `name` property and it seems to work in Chrome. I might have messed the check itself though. Could you check again?

AkeluX:
You are absolutely right (as always). Could you then revert my changes to `instanceof`?

PR #767, opened 2018-06-29
Refactored Standalone example as the Offline app

Beak-man:
### Description of the Change
Unlike in previous times, the Standalone example wasn't working right off the bat after executing our build process. This was simply due to the outdated reference in the script tag to `worldwind.min.js`. It's fixed now.

Moreover, this PR is proposing a name change to the example in order to better convey what it's supposed to do. While the name 'Standalone' was appropriate, it's maybe too open to interpretation. Calling the example 'Offline' sounds more succinct, particularly for our users that are interested in such functionality.

It was also considered if this example sit well among the others, since it's used as a tutorial. For now, we're proposing bundling it alongside the Apps.

Some comments were modified as well.

### Potential Drawbacks
- While renaming the example seems fitting, we're still using the previous 'standalone' moniker for the data bundle itself and its tar.gz file (and uncompressed folder). Do you think it would be worthwhile to rename it to align it with this renaming, @pdavidc?

- I'm not too sure if this sits well among the Apps, either. The idea is to use it only as a tutorial, but we need to maintain its code somewhere in the repository. 

### Applicable Issues
Relates to #539.


pdavidc:
@Beak-man @eirizarry I think this is the right move for the Standalone example. _Offline_ is indeed a better name, for all the reasons you mention.

1. We can certainly rename the data package, though I'm concerned that the package gives the false impression that this is the _only_ data that can be viewed offline. Do you think it is important to have a single package to download (that also matches the tutorial name)? If not, would it be an improvement to package up each example dataset separately (Blue Marble, Landsat, Elevations)? 

2. Put the code wherever you think it belongs. Are we expecting developers to run the code? If so, where from? (dev environment?, online?)

3. The reference to worldwind.min.js was intentional, as the example is only expected to run online. That said, this is in retrospect an odd decision given that the code is in examples, and we certainly don't need to stick with it. How we reference the WorldWind library depends on the answer to #2.

Beak-man:
@pdavidc: 

1. Separately packaging each dataset sounds like a reasonable improvement. This way we avoid the apparent pigeonholing of the offline data sources available for WebWW. Plus we won't be needing to keep constant maintenance on the package name if we decide to change the name of the example again.

2. @eirizarry and I considered this. I'm of the thinking that this example should be ran, well, offline. (With an online tutorial that describes how to use it and what the results will be, through screenshots).

3. Thanks for the explanation. Indeed this depends on the previous point. What is your opinion, @eirizarry?

Beak-man:
Hi @markpet49. I included a change on the SimplestExample to reflect the location of the WorldWind library given our current build process. Consider if you like the idea of renaming the Standalone example into the Offline app.

markpet49 (changes requested):
@Beak-man Hi Miguel, let's change the name of the run configuration to 'apps/Offline' so that it's consistent with the other run configurations. (e.g. apps/Collada, etc.)

markpet49 (approved):

barionleg:
standalone data link https://github.com/aibolem/WebWorldWind/releases/tag/v.0.12.0

1.8 gb

image

[NASA family = С₽uтнuი©$ÅРЬ 🛰 мультиплес Спутни©а* (натеслაвеßи)
](https://github.com/barionleg/ABC_ub_particles/wiki)

matt-msi:
> standalone data link https://github.com/aibolem/WebWorldWind/releases/tag/v.0.12.0
> 
> 1.8 gb
> 
> image
> 
> [NASA family = С₽uтнuი©$ÅРЬ 🛰 мультиплес Спутни©а* (натеслაвеßи) ](https://github.com/barionleg/ABC_ub_particles/wiki)

Standalone data at this link appears to be corrupted.  Is the data mirrored anywhere else?

PR #769, opened 2018-07-06
Track WebGL extensions per DrawContext

### LINKED ISSUE #768: Track GL extensions in DrawContext
The proposition is to track GL extensions in the DrawContext as explained in #766. @strikerM please propose a concrete solution for this as part of a separate pull request.

Child of #631

strikerM:
### Description of the Change
Provides a method for renderables to request gl extensions in  a way that works in a multi window setup, context lost scenarios and minimizes ```gl.getExtension``` calls.

The extensions are cached on the drawContext instance, repeated calls for the same extension will not make a gl call. 

The cache is cleared on context lost and context restored.

Example:
```javascript
var extension = dc.getExtension("OES_element_index_uint");
```
The method returns the gl extension or null if the extension is not available.

### Potential Drawbacks
None

### Applicable Issues
Closes #768


src/render/DrawContext.js
```diff
@@ -1573,5 +1582,34 @@ define([
...
+                    "missing extensionName"));
+            }
+
+            if (typeof extensionName !== "string") {
+                throw new ArgumentError(Logger.logMessage(Logger.LEVEL_SEVERE, "DrawContext", "getExtension",
+                    "extensionName is not a string"));
+            }
+
+            if (extensionName in this._glExtensionsCache) {
+                return this._glExtensionsCache[extensionName];
```
AkeluX: This could be simplified a bit by using the opposite condition and saying: if not in the cache, try to load it. Then you always return the value from the cache. What do you think?

src/render/DrawContext.js
```diff
@@ -419,6 +419,13 @@ define([
 
             // Intentionally not documented.
             this.pixelSizeOffset = 0;
+
+            /**
+             * A place to store gl extensions for this drawContext.
+             * @type {Object}
+             * @private
+             */
+            this._glExtensionsCache = Object.create(null);
```
AkeluX: This notation with `_` and `@private` is new to this class. For consistency, I would use the same approach as above for `pixelSizeOffset`. Or is this the new convention @pdavidc ?

src/render/DrawContext.js
```diff
@@ -419,6 +419,9 @@ define([
 
             // Intentionally not documented.
             this.pixelSizeOffset = 0;
+
+            // Intentionally not documented.
+            this.glExtensionsCache = Object.create(null);
```
AkeluX: @strikerM, we were briefly looking at the differences between `Object.create(null)` and `{}`. They are pretty clear with respect to the resulting objects. However, have you made any assessment of the performance gain? More generally, was there something that drove your decision to use `Object.create(null)` here? We would like to understand whether this is a pattern that should be promoted in other places of the framework.


For reference, see the validated response at https://stackoverflow.com/questions/32262809/is-it-bad-practice-to-use-object-createnull-versus

src/render/DrawContext.js
```diff
@@ -1573,5 +1578,32 @@ define([
...
+         * A wrapper around gl.getExtension method, that also caches the extensions.
+         *
+         * @param {String} extensionName The name of the WebGL extension.
+         *
+         * @returns {Object|null} A WebGL extension object, or null if the extension is not available
+         *
+         * @throws {ArgumentError} if the name is missing or not a string
+         */
+        DrawContext.prototype.getExtension = function (extensionName) {
+            if (!extensionName) {
```
AkeluX: This method being used internally only, we think that the checks of `extensionName` are superfluous. If you concur, could you remove them?

AkeluX: As discussed, there are checks in internal methods throughout the framework, see https://github.com/NASAWorldWind/WebWorldWind/blob/develop/src/cache/GpuResourceCache.js#L142 for example.

Please remove only the type check and leave the null check.

src/render/DrawContext.js
AkeluX: As discussed, the performance gain is neglectable in this case. This is however a better pattern for showing that this is a data store (dictionary) and not an object.

PR #770, opened 2018-07-07
Bug/jszip require version

balhar-jakub:
### Description of the Change

Update to the JsZip library so that it is compatible with webpack build and AMD modules. 

Because otherwise the AMD version of the library breaks. 

### Applicable Issues


AkeluX:
@emxsys could you confirm that this solves the issue in your app? We couldn't test your particular case, but have validated this with another app already.

pdavidc:
@AkeluX @balhar-jakub Since you've already validated these changes, let's merge them now.
@emxsys may not be able to test the fix in his environment soon, and the changes are simple.

PR #773, opened 2018-07-13
Updated BlueMarble and Landsat layers to capture their full resolution

### LINKED ISSUE #772: Enhance BlueMarble and Landsat layer resolution
WebWorldWind's built-in Blue Marble and Landsat image layers capture a majority of each dataset's respective original, but don't capture the full resolution. Enhance both layers to ensure the underlying dataset's maximum resolution is captured by the WorldWind layer.

pdavidc:
- Blue Marble's native resolution is 4.166666667E-03 deg/pixel.
- Landsat's native resolution is 1.38889E-04 deg/pixel.
- Increased BMNGLayer's maximum resolution from 5.5E-0.3 deg/pixel to 1.4E-0.3 deg/pixel.
- Increased BMNGLandsatLayer's maximum resolution from 1.7E-0.4 deg/pixel to 4.3E-0.5 deg/pixel.
- Given the modest increase in resolution, this change has no effect on runtime performance.

Closes #772

### DIFF
--- a/src/layer/BMNGLandsatLayer.js
+++ b/src/layer/BMNGLandsatLayer.js
@@ -37,8 +37,9 @@ define([
          * @classdesc Displays a combined Blue Marble and Landsat image layer that spans the entire globe.
          */
         var BMNGLandsatLayer = function () {
+            // This LevelSet configuration captures the Landsat resolution of 1.38889E-04 degrees/pixel
             TiledImageLayer.call(this,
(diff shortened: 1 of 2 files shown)


Beak-man:
The increase in resolution is immediately resolved. Blue Marble is on the left, Landsat on the right. Both views are at 6 km altitude.

davepr

With right click -> _Open image in new tab_ this image can be seen in high resolution. Please note that this GIF induced some banding not present in the live scene.

AkeluX:
Wow, that makes quite a difference. Very nice!

PR #775, opened 2018-07-15
Clarified console messages relating to WebGL support

pdavidc:
Changed the console message emitted during WorldWindow construction when a WebGL context cannot be created. These messages appear when the browser does not support WebGL, or when WebGL support has been disabled (either by the user, or by a group policy).

Old console message:

webgl message before

New console message:

webgl message after


pdavidc:
@AkeluX @markpet49 We're encountering browsers on corporate networks that have WebGL disabled by a group policy. The intention of this change is to make the problem clear to developers affected by such a group policy.

Please let me know if you have any questions or concerns. If not, your timely approval of these changes would be very helpful.

PR #776, opened 2018-07-17
Add getReferencePosition and moveTo to SurfaceSector

claudiaifrim:
### Description of the Change
Add moveTo and getReferencePosition methods for SurfaceSector.

Methods used in Shape editor.

### DIFF
--- a/src/shapes/SurfaceSector.js
+++ b/src/shapes/SurfaceSector.js
@@ -120,16 +120,20 @@ define([
 
         // Internal use only. Intentionally not documented.
         SurfaceSector.prototype.moveTo = function (globe, position) {
-            if (this._boundaries.length > 0) {
-                this._boundaries = this.computeShiftedLocations(globe, this.getReferencePosition(), position,
-                    this._boundaries);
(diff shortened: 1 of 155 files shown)


src/shapes/SurfaceSector.js
```diff
@@ -113,5 +113,29 @@ define([
             this._boundaries[3] = new Location(sector.minLatitude, sector.maxLongitude);
         };
 
+        // Internal use only. Intentionally not documented.
+        SurfaceSector.prototype.getReferencePosition = function () {
+            // Assign the first position as the reference position.
```
AkeluX: This comment is inaccurate. The centroid of the sector is used.

src/shapes/SurfaceSector.js
```diff
@@ -113,5 +113,29 @@ define([
...
+            if (this._boundaries.length > 0){
```
AkeluX: The check, if any, should be on the existence of the sector that is used afterwards. Are there cases where this sector can be `null` or `undefined`?

PR #784, opened 2018-08-16
Suppress automated API doc deployment

pdavidc:
Temporarily suppress automated API doc deployment to GitHub Pages until the root cause of build errors can be identified and resolved.


Beak-man:
Looks good.

PR #797, opened 2019-01-29
Add an option to set the number of levels in a heatmap

balhar-jakub:
### Description of the Change

Provide configuration for the number of levels for the HeatMap. Using the one level the behaviour of the HeatMap is similar to the Analytical Surfaces from the World Wind Java.

The HeatMap Layer implementation was incorrectly applying the opacity to the coloured tiles. Originally the opacity for black and white layer was applied in the same way to the coloured layer. Now there is a possibility either to keep this behaviour or change the final opacity based on the resulting opacity. 

The functionality added to the HeatMap provides higher flexibility to the users of this functionality with very little in terms of added complexity in the code. 

### Benefits

Added support for the behaviour of the HeatMap similar to the one with the Analytical Surfaces. 

Support for the different opacity in the coloured layer.


src/layer/heatmap/HeatMapLayer.js
```diff
@@ -51,12 +51,18 @@ define([
      * @classdesc A HeatMap layer for visualising an array of measured locations.
      * @param {String} displayName This layer's display name.
      * @param {MeasuredLocation[]} measuredLocations An array of locations with measures to visualise.
+     * @param {Number} amountOfLevels Optional. If provided it specifies the amount of levels that will be generated for
```
AkeluX: `TileImageLayer` calls this property `numLevels`. Please use the same name for consistency.

src/layer/heatmap/HeatMapLayer.js
```diff
@@ -51,12 +51,18 @@ define([
...
+     *  this layer. If you have only one, it behaves in a similar way to the Analytical surface from the World Wind Java.
+     *  If you have multiple levels then whenever the user zooms over certain threshold, the new version of the layer
+     *  will be generated.
+     * @param {Number} opacity Optional. The final opacity of this layer. Number between 0 and 1. If the opacity isn't
```
AkeluX: What is the use case for the default situation (value undefined)? I suggest having full opacity as default and allow setting a lower one via this optional property.

In addition, as it is optional, I would also have this option as a property and not an argument of the constructor.

balhar-jakub:
Thanks for the feedback. The numLevels were changed and opacity moved to the property. As for the default options:

The current version of the HeatMap. This is what I refer to as the default value. The colour and opacity go hand in hand with the variant from the canvas. 
image

The opacity 1 version of the current implementation. 
image

The opacity 0.5 version of the current implementation. 
image

It might be interesting to keep the default behaviour, but it is true, that it shouldn't be the default behaviour. For now, I provided the update without this "default" version. The default is now 1.

src/layer/heatmap/HeatMapColoredTile.js
```diff
@@ -33,13 +33,16 @@ define([
      *  color strings.
      *  @param options.extendedWidth {Number} Optional. Minimal width that needs to be retrieved for colorization.
      *  @param options.extendedHeight {Number} Optional. Minimal height that needs to be retrieved for colorization.
+     *  @param options.opacity {Number} Optional. The opacity applied to the calculated colors. The opacity is between 0 and 1. If this value isn't
```
AkeluX: Please specify what 0 and 1 are in terms of transparency. In addition, the default is not longer the one mentioned.

src/layer/heatmap/HeatMapColoredTile.js
```diff
@@ -33,13 +33,16 @@ define([
...
+     *  @param options.opacity {Number} Optional. The opacity applied to the calculated colors. The opacity is between 
+     *      0 (fully transparent) and 1 (fully opaque). The default value is 1.
      */
     var HeatMapColoredTile = function(data, options) {
         HeatMapTile.call(this, data, options);
 
         this._extendedWidth = options.extendedWidth;
         this._extendedHeight = options.extendedHeight;
         this._gradient = this.gradient(options.intensityGradient);
+        this._opacity = options.opacity * 255 || 1;
```
AkeluX: If `options.opacity` equals 0, the resulting opacity will be 1 with this change. Not what you wanted I guess.

balhar-jakub:
While testing with undefined and null as the opacity for the layer specified on the layer it turned out that there was some interference from the opacity property on the Layer. 

Therefore I had to rename the property. The current name is tileOpacity.

balhar-jakub:
At the moment the opacity property, as well as tileOpacity property, was removed from the HeatMap. The standard opacity mechanism implemented in the TiledImageLayer and SurfaceTileRendererProgram is used. Therefore setting the opacity property on the HeatMapLayer is correctly used. If set to undefined or null the layer won't be visible. This is the default behaviour of the mechanism. 

It is possible still to change the amount of the levels for the HeatMapLayer. The default is 18. 0 is invalid value and as such is replaced by the 18 the same way as undefined, null and other falsy values. 

We aren't testing for the user providing value below zero. The layer just won't work in such case.

PR #826, opened 2020-06-04
Fix to correctly display KML files with no altitudeMode specified

Beak-man:
### Description of the Change
Corrected behavior of KML file rendering when the file does not specify an altitude mode.
As per KML's 2.2 spec, this should fallback to a 'clampToGround' mode, which WebWW should internally recreate as a 'SurfacePolygon'. Previously, WebWW was assuming that 'altitudeMode' was always defined, causing a 'Polygon' to be used instead, causing clipping with the terrain.

The check for no altitudeMode is now correctly fulfilled, defaulting behavior to 'clampToGround' and invoking a 'SurfacePolygon' shape to be rendered.

Npm packages have been updated as well.

### DIFF
--- a/src/formats/kml/geom/KmlPolygon.js
+++ b/src/formats/kml/geom/KmlPolygon.js
@@ -140,11 +140,13 @@ define([
      * @param styles.highlight {KmlStyle} Style to apply when item is highlighted. Currently ignored.
      */
     KmlPolygon.prototype.createPolygon = function(styles, fileCache) {
+        console.log(this.kmlInnerBoundary && this.kmlInnerBoundary.kmlAltitudeMode === WorldWind.CLAMP_TO_GROUND);
         if(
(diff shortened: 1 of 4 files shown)


src/formats/kml/geom/KmlPolygon.js
```diff
@@ -206,6 +211,16 @@ define([
         return locations;
     };
 
+    /**
+     * @inheritDoc
+     */
+    KmlPolygon.prototype.isValidAltitudeMode = function (altMode) {
```
markpet49, line 217: Lets do this like this:
    KmlPolygon.prototype.isValidAltitudeMode = function (altMode) {
        return WorldWind.CLAMP_TO_GROUND === altMode
            || WorldWind.RELATIVE_TO_GROUND === altMode
            || WorldWind.ABSOLUTE === altMode;

    };

PR #837, opened 2020-09-16
Retrievable pick point

markpet49:
Adds the ability to determine where a user has clicked on a 3D model in a WorldWind Scene. Also fixes a minor bug in PlacemarkAttributes: When cloning another set of attributes, some sub-attributes were being referenced directly instead of cloned themselves.

### DIFF
--- a/examples/Collada.js
+++ b/examples/Collada.js
@@ -32,49 +32,80 @@
 requirejs(['./WorldWindShim',
         './LayerManager'],
     function (WorldWind, LayerManager) {
-    "use strict";
+        "use strict";
 
         // Tell WorldWind to log only warnings and errors.
-    WorldWind.Logger.setLoggingLevel(WorldWind.Logger.LEVEL_WARNING);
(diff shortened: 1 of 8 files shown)


Beak-man:
Hi @markpet49 The intersection code works excellently. It accounts for multiple intersections within concavities. I added some comments to the COLLADA example to make it more of a tutorial for WebWW newcomers, putting it in line with our other examples. 

I'm torn, however, on adding the intersection example to the COLLADA example, which is already somewhat convoluted due to the asynchronous nature of the COLLADA loader. It may be best to add this to a separate example, or to our more fully-featured COLLADA reference application in here:

https://github.com/NASAWorldWind/WebWorldWind/blob/develop/apps/Collada/Collada.js

In the latter case, it would bring the added benefit of showing the intersection feature with multiple models, serving as a more comprehensive testing case. What do you think?

Beak-man (approved):
Looks good! We'll move the intersection example to the COLLADA app at a later point.

PR #838, opened 2020-09-18
Added 'prepare' npm script

Beak-man:
### Description of the Change
Added `prepare` script for the benefit of developers that pull WebWW's repo from GitHub as dependency.

Reference: https://docs.npmjs.com/misc/scripts#prepublish-and-prepare


markpet49 (approved):
Looks good

PR #863, opened 2021-11-18
Update offline example

Beak-man:
### Description of the Change
- Updated URL of data package download.
- Reshuffled code to make it more sequential in order to improve readability.
- Modified comments regarding the naming of the sources as "REST". Used "local" instead, since the example is supposed to convey the usage of WebWW in a locally executed web server.


markpet49 (approved):
Looks good.

Issue #834, opened 2020-08-17
Misleading naming in GeoTiffReader.js 


Cazadorro:
I would submit a pull request myself, but I only encountered these issues helping someone else who uses this library, I don't use NWW myself at the moment, and don't understand the javascript build environment. 

Recently I was helping debug someone's code that used the latest minified NWW release as of 8/17/2020. 
This line https://github.com/NASAWorldWind/WebWorldWind/blob/20a5a0cce09bde159cb64d3a5136190fc4d52e6c/src/formats/geotiff/GeoTiffReader.js#L203 looked like dead code to us.  The naming of "get" is misleading.  It's internal, but it was necessary for us to look into the internals of the project in order to figure out what our issue was with loading a geotiff. 

it appears that https://github.com/NASAWorldWind/WebWorldWind/blob/20a5a0cce09bde159cb64d3a5136190fc4d52e6c/src/formats/geotiff/GeoTiffReader.js#L219 initializes the endianness after the file has been read, using the first two bytes with a known value in order to determine the order of the data.  I would suggest something like "initEndiannessFromFile" so that there's no confusion, but I don't really mind what it is as long as it doesn't insinuate it is a read only function.

Beak-man:
Hi @Cazadorro, thanks for your feedback. After doing a review, we included the change you suggested in PR #864

PR #864, opened 2022-01-07
Renamed getEndianness function to better convey its behavior

Beak-man:
### Description of the Change

Renamed `getEndianness()` function to `initEndiannessFromFile()` since the original name conveys that it's a read-only function, when it's not.

Addresses #834


markpet49 (approved):
Looks good.

PR #871, opened 2022-04-12
Vscode conversion

markpet49:
### Description of the Change

Fixes the calculation for tile splitting when the viewpoint is close to the surface allowing the highest resolution tiles to be loaded at those altitudes. Also adds support for editing/debugging with VSCode. 

Improves the map view at low viewpoint altitudes.

### Benefits

Improves the map view at low viewpoint altitudes.

### Potential Drawbacks

None known


Beak-man (approved):
That was quite the bugfix @markpet49. Great job!
Before and after screenshots:
BingMapsBugfix

Beak-man:
@markpet49 a unit test of the `Sector` class is failing after the latest commits:
image

Beak-man (approved):
Unit tests are now passing in both headless browsers, and the npm package is assembling itself properly.

PR #875, opened 2022-04-29
Fixes for launch.json and 2D projection related calculations

markpet49:
### Description of the Change
Fixes the problem with 2D projections that was introduced in a previous pull request.


Beak-man (approved):
Looks good!

PR #876, opened 2022-04-30
Fix for picking transparent placemarks

markpet49:
### Description of the Change
Shader fix for picking placemarks and other textures that are mostly transparent. Includes a test case in the PlacemarksAndPicking example.


PJHogan:
Wahoo!

PJHogan:
If Unblocked!!!

PJHogan:
Ooch! Ouch!

Beak-man (approved):
Looks good!

PR #877, opened 2022-05-10
Reverted WKT coordinate parsing to first read latitude and then longitude

Beak-man:
### Description of the Change

- Changed the way `WktObject` reads coordinates from long-lat to lat-long.
- Adjusted the WKT unit test accordingly.
- Modified WKT example styling to have the WKT string entry textbox and "Show" button to be visually in line with the Destination textbox.

Tested proper functioning of WKT example and the `WktTokens` unit test.

In #780 the order of geographical coordinates was changed to be read with longitude before latitude for WKT strings.

This introduced errors in the Well-Known Text example:

- The `MULTIPOLYGON` green triangles are no longer displayed.
- Custom WKT entry no longer works.

It seems that there's ambiguity on the precedence of latitude or longitude in WKT. OpenGeospatial's specification on Simple Feature Access, section 7 [[ > ](https://portal.ogc.org/files/?artifact_id=25355)], which is mentioned as the reference spec for our format implementation in `src/formats/wkt/README.MD`, doesn't state any precedence either for latitude and longitude.

The following discussion provides what is the most reasonable course of action in my estimation:

> For the axis order for any coordinate reference system where EPSG is the authority, you should use their registry https://epsg.org/home.html

> For example for epsg:4326 (https://epsg.org/crs/wkt/id/4326).
> The WKT output shows that for EPSG:4326 the axis order is lat/long

````
GEODCRS["WGS 84",
  DATUM["World Geodetic System 1984",
    ELLIPSOID["WGS 84",6378137,298.257223563,LENGTHUNIT["metre",1.0]]],
  CS[ellipsoidal,2],
    AXIS["latitude",north,ORDER[1]],
    AXIS["longitude",east,ORDER[2]],
    ANGLEUNIT["degree",0.01745329252],
  ID["EPSG",4326]]
````

> Most EPSG geographic geographic coordinate systems follow the latitude then longitude, coordinate axis order, following the default of ISO 6709 Standard representation of geographic point location by coordinates

https://gis.stackexchange.com/questions/366853/wkt-coordinate-inner-order-long-lat

This change aligns WKT parsing with the ISO standard, making the WKT example to properly work again.

### Benefits

The WKT example works as intended. WKT parsing is now in compliance with ISO 6709 [ [>](https://www.iso.org/standard/39242.html) ], and in line with other geographical string formats within the library.

### Potential Drawbacks

There seems to be ambiguity on how WKT parsing should work. Other products do give precedence to longitude before latitude, but in my estimation, it's appropriate that we align with the ISO standard, considering that the OpenGeospatial spec is ambiguous in regards to how WKT should be parsed if CRS is not provided.

### DIFF
--- a/examples/WKT.html
+++ b/examples/WKT.html
@@ -35,10 +35,12 @@
                     </button>
                 </span>
             </div>
-            <h4>WKT</h4>
-            <div>
-                <input type="text" id="wkt" />
-                <input type="button" id="showWkt" value="Show" />
+            <h4>Custom WKT entry</h4>
+            <div class="input-group">
+                <input type="text" class="form-control" id="wkt" />
+                <span class="input-group-btn">
+                    <input type="button" class="btn btn-primary" id="showWkt" value="Show" />
+                </span>
             </div>
         </div>
         <div class="col-sm-9" id="globe">
--- a/examples/WKT.html
+++ b/examples/WKT.html
@@ -39,7 +39,7 @@
             <div class="input-group">
                 <input type="text" class="form-control" id="wkt" />
                 <span class="input-group-btn">
-                    <input type="button" class="btn btn-primary" id="showWkt" value="Show" />
+                    <input id="showWkt" class="btn btn-primary" type="button" value="Show"/>
                 </span>
             </div>
         </div>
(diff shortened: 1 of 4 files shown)


markpet49 (approved):
Looks good, thanks for the thorough research.

PR #879, opened 2022-05-16
Sentinel WMTS outdated API key removal

Beak-man:
- Outdated API key was removed.
- Modified documentation for clarity.
- After testing with new API key kindly provided to us by [Sentinel Hub](https://www.sentinel-hub.com/), it's decided that we should return to review this example application at a later point.


markpet49 (approved):
Looks good.

PR #896, opened 2023-10-24
Added version consistency check before each build

Beak-man:
### Description of the Change
Additionally to the version number located in package.json which provides us with versioning for the npm package, we internally store a VERSION property in the main WorldWind object. Since the latter predated the publishing of the npm package and due requiring manual updating, it has not been in sync with the package.json value.

- This change adds a script that verifies consistency between these two values before performing a build, canceling the build if the check fails. 
- The value in WorldWind.js is updated to reflect the numbering as found in the npm package and GitHub's releases.

Tested regex in line 23 of the script at https://regex101.com/. I also tested an assortment of version numbers and nonsensical strings.

It provides the developer the means to ensure that version numbers are consistent across the project without incurring in breaking changes to the WorldWind object.

### Benefits
Simple way to maintain version consistency between the two files where the value is stored.

### Potential Drawbacks
For future releases, It still requires manually updating the version number in WorldWind.VERSION. Since automatically rewriting it with each build could be considered an antipattern, manual updating is deemed acceptable.

### DIFF
--- a/package.json
+++ b/package.json
@@ -17,7 +17,8 @@
     "url": "https://github.com/NASAWorldWind/WebWorldWind.git"
   },
   "scripts": {
-    "build": "grunt",
+    "build": "npm run check-version && grunt",
+    "check-version": "node tools/versioncheck.js",
     "clean": "grunt clean",
     "doc": "grunt jsdoc",
     "prepare": "npm run build",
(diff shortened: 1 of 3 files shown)


ApacheAde:
Version

ApacheAde (changes requested):
DALL·E 2023-10-24 21 17 08 - Re imagine the Friends TVs show cast as snoopy peanuts characters

markpet49 (changes requested):
I can't find any evidence that versioncheck.js gets executed with the build command 'npm run build'. How does versioncheck get initiated?

PJHogan (approved):

markpet49 (approved):
Looks good.

